PR #2, opened 2019-02-08
Initial project set up

rebecabkerr:
**An initial project set up including:**
react, typescript, redux, saga, patternfly 4 and patterfly-react 4, SASS

**Initial commit includes the following Pages:**
- Home
- Not found, 404 page
- Login page: Note: this is only displaying the form, submit, message and error handlers will be done at a later date

**Notes:** 
- So far, there are 2 main page layout: full width (without sidebar or header, like Login page) and Vertical Page layout (include sidebar and header, Dashboard)
- Patterfly 4 scss library is imported in through assets/scss/vendors/patternfly. This will be slowly cleaned up to include only the assets used in the app. 
- PF4 specific variables may be modified as neede in assets/scss/helpers/_pf4-variables.scss

**Some supporting initial features include:**
- Global Sass styles - (/assets/scss/main.scss)
- Component specific style (example: Pages/LogIn/login.scss)
- Redux store: only the uil state manager is currently being used. The message store is a stub for future feature development


jcaianirh:
@rebecabkerr for the components, can we name the files {componentName}.tsx instead of index?

rebecabkerr:
Component names changed to {ComponentName}.tsx - commit 810d889

src/store/ui/types.ts
```diff
@@ -0,0 +1,32 @@
...
+*  File:            ui/types.ts
+*  Description:     Holds types and constants for managing global UI states
+*  Author:          ChRIS ui Demo                 
+*/
+
+
+import keyMirror from "keymirror";
+
+// Description: state for UI Manager
+export interface IUiState {
```
jcaianirh: For UI component state, can we keep the state in the component, and leave the app wide states here? Is there a benefit of tracking ui state in redux?

rebecabkerr: My preference is to keep state that is used in more than one component centralized in the redux store. There is really no "right" answer for this specific question, just best practice and dev preferences (Ref: https://redux.js.org/faq/organizing-state#do-i-have-to-put-all-my-state-into-redux-should-i-ever-use-reacts-setstate). Generally, the only time I use local state is in things like form components. Otherwise, my preference is to have it centralized in the store.

jcaianirh: I think my preference would be otherwise, but lets go with what you have for now.

jcaianirh:
@rebecabkerr I installed and everything ran according to your readme. Great job on this kickoff pr for the ui!!

PR #11, opened 2019-03-22
Render Charts

gagan-s-kumar:
PR is for rendering a Line chart and a bar chart.
The rendering will happen in the Dashboard but it can be placed anywhere since it is contained in a component.

Components use c3js library to render the charts. cs3js renders the charts based on the JSON object passed as the input from the component.


src/pages/Dashboard/Dashboard.tsx
```diff
@@ -17,15 +19,10 @@ class DashboardPage extends React.Component<AllProps> {
...
-          <div>
-            <h1 className="pf-u-mb-md">Welcome to ChRIS UI site's Dashboard</h1>
-              <Alert
-                aria-label="welcome wagon"
-                variant="info"
-                title="Welcome!"  >
-                Welcome to ChRIS UI site's Dashboard
-              </Alert>
-          </div>
+          {/* UnComment to render
```
jcaianirh: do we need this comment?

src/components/chart/SegmentAnalysis.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  setFilter(){
+    var filteredData : any[] = [];
+    var parsedData : any[] = [];
+    if(this.state.pushedSegments.length > 0) {
+        filteredData = this.state.pushedSegments.map(segment =>
+          this.getSegmentData(segment));
+    }
+    parsedData = this.parseData(filteredData);
+    // return parseData
```
jcaianirh: remove comment

src/components/chart/SegmentAnalysis.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+    this.changeData = this.changeData.bind(this);
+  }
+
+  componentDidMount() {
+    this.callChart(defaultChartData, defaultSegments);
+  }
+
+  callChart(inputChart: any, segments: any) {
+    var chart = c3.generate({
```
jcaianirh: I think we can get rid of this var statement, by adding chart to state instead of having this global var. maybe return the results of c3.generate in the callChart function, and have the caller of this function add the results to the state variable called chart.

src/components/chart/VolumeGrowth.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    this.changeData = this.changeData.bind(this);
+  }
+
+  componentDidMount() {
+    this.callChart(defaultChartData);
+  }
+
+  callChart(inputChart: any) {
+    var chart = c3.generate({
```
jcaianirh: same comment as above. let me know if that works for you

src/pages/Dashboard/Dashboard.tsx
gagansk: Since the component will be moved from the Dashboard to somewhere else, I thought it will be easier to understand. I have removed the comment in the commit.

src/components/chart/SegmentAnalysis.tsx
gagansk: Removed.

src/components/chart/SegmentAnalysis.tsx
gagansk: The var statement is not global. It will just retain the result of the render operation. By default, c3.generate will bind the rendering function to div with id '**chart**'. I have now, exclusively bound the rendering to the component with the corresponding div.

src/components/chart/VolumeGrowth.tsx
gagansk: The comment has been addressed above.

gagansk:
Thank you for the comments, @jcaianirh. By default, rending by c3js is bound to div id 'chart'. Now, I have changed to the respective component names. However, I don't think introducing chart in the state will make any difference since rendering will be called only twice. Once when the Component gets mounted and the other time is when there is a change in the filters. 

Please let me know if your comments haven't been addressed.

jcaianirh:
I would make App.test.tsx work by changing it to this:
```import React from 'react';
import ReactDOM from 'react-dom';
import Main from './main';
import {createBrowserHistory} from "history";
import configureStore from "./store/configureStore";

it('renders without crashing', () => {
  const div = document.createElement('div');
  const history = createBrowserHistory();
  const store = configureStore(history);
  ReactDOM.render(<Main store={store} history={history} />, div);
  ReactDOM.unmountComponentAtNode(div);
});```

File: src/components/chart/SegmentAnalysis.tsx (hunk 3 above)
Comment (jcaianirh) [reply src/components/chart/SegmentAnalysis.tsx]: @gagan-s-kumar chart is unnecessary here. please remove.

File: src/components/chart/SegmentAnalysis.tsx [hunk 5]
```diff
@@ -0,0 +1,145 @@
...
+      },
+      padding: {
+        top: 30,
+        bottom: 20
+      }
+    });
+  }
+
+  parseData(filteredData: any) {
+    var leftHemisphereData = ['leftHemisphere'];
```
jcaianirh: let leftHemisphereData instead of var for both variables since you are making changes to both of them.

src/components/chart/SegmentAnalysis.tsx
```diff
@@ -0,0 +1,145 @@
...
+    // Parse for the leftHemisphereData and rightHemisphereData
+    for(var i = 0; i < filteredData.length; i++){
+      leftHemisphereData.push(filteredData[i][1]);
+      rightHemisphereData.push(filteredData[i][2]);
+    }
+    return [leftHemisphereData, rightHemisphereData];
+  }
+
+  getSegmentData(segment: any){
+    var segmentData;
```
jcaianirh: let segmentData here.

src/components/chart/SegmentAnalysis.tsx
```diff
@@ -0,0 +1,145 @@
...
+    segmentData = csvData.find(function(segmentData) {
+      return (segmentData[0] === segment);
+    });
+    return segmentData;
+  }
+
+  setFilter(){
+    var filteredData : any[] = [];
```
jcaianirh: let instead of var

src/components/chart/SegmentAnalysis.tsx
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  changeData(selectedSegments: any) {
+    // Call back function to avoid asynchronous setState
+    var processedData;
+    this.setState({
+      pushedSegments : selectedSegments
+    }, () => {
+      //Input processing
+      processedData = this.setFilter();
```
jcaianirh: const pocessedData = this.setFilter(); instead of var

src/components/chart/VolumeGrowth.tsx
```diff
@@ -0,0 +1,156 @@
...
+      tooltip: {
+        format: {
+            title: function (d) { return d + ' Months old'; },
+        }
+      }
+    });
+  }
+
+  getSegmentData(segment: string){
+    var segmentData = chartData.find(function(segmentData) {
```
jcaianirh: for this file, use lets and const similar to the other graphing component.

src/components/chart/SegmentAnalysis.tsx
gagansk: var chart is removed

src/components/chart/SegmentAnalysis.tsx
gagansk: var is replaced by let

src/components/chart/SegmentAnalysis.tsx
gagansk: Changed to const since the value wouldn't change

src/components/chart/SegmentAnalysis.tsx
gagansk: let used instead of var

src/components/chart/SegmentAnalysis.tsx
gagansk: let has replaced var

src/components/chart/VolumeGrowth.tsx
gagansk: Done

gagansk:
@jcaianirh, as suggested, vars have been replaced by lets if they are changing in the function scope or const if they are not changing.

PR #16, opened 2019-04-02
Unit test for the feed reducer.

ethanhou99:
There are two types of errors:
1. For the __tests__/store/plugin/action.test.ts:
```
SyntaxError: /Users/ethanhou/Desktop/Test/ChRIS_ui/src/api/models/file-explorer.ts: Missing class properties transform.
      22 | export default class UITreeNodeModel {
      23 |   private _items: IFeedFile[];
    > 24 |   private _worker: IUITreeNode = {
      25 |     module: "root",
      26 |     children: []
      27 |   };
```
2. For the tests__/store/feed/reducer.test.ts
```
TypeError: Cannot read property 'results' of undefined

      25 |       // Note: when using the ChrisAPI the items will be action.payload.collection.items
      26 |       // return { ...state, items: action.payload.collection.items }; //Note: For API call... stub
    > 27 |       return { ...state, items: action.payload.data.results };
         |                                                     ^
      28 |     }
      29 |     case FeedActionTypes.FETCH_REQUEST: {
      30 |       return { ...state };
      at results (src/store/feed/reducer.ts:27:53)
      at Object.<anonymous> (__tests__/store/feed/reducer.test.ts:82:7)
```

### DIFF
--- a/__tests__/store/feed/reducer.test.ts
+++ b/__tests__/store/feed/reducer.test.ts
@@ -1,0 +1,157 @@
+import React from 'react';
+import { feedReducer } from '../../../src/store/feed/reducer'
+import { IFeedState, FeedActionTypes } from "../../../src/store/feed/types";
+import { any } from 'prop-types';
+import { IFeedItem } from "../../../src/api/models/feed.model";
+import { IPluginItem } from "../../../src/api/models/pluginInstance.model";
+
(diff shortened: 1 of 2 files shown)


ethanhou99:
Some bugs need to be fixed.

PR #18, opened 2019-04-03
update reducer of redux message and ui

xhyzzZ:


src/__tests__/store/message/reducer.test.ts
```diff
@@ -0,0 +1,74 @@
+import { messageActionTypes, IMessageState } from "../../../src/store/message/types";
```
jcaianirh: imports not correctly finding files in path. fix paths in all import statements.

src/__tests__/store/message/reducer.test.ts
```diff
@@ -0,0 +1,74 @@
...
+import { messageReducer } from "../../../src/store/message/reducer";
+
+const initialState: IMessageState = {
+    message: undefined,
+    confirmation: undefined
+};
+const testState : IMessageState = {
```
jcaianirh: The initial state and testing state shouldn't be the same, right? We want to actually test that something is happening rather than just passing undefined around. update the tests to actually test for sample message and confirmation strings please in all tests.

src/__tests__/store/ui/reducer.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    it('should return the initial state', () => {
+        expect(uiReducer(undefined, {
+            type: null
+        })).toEqual(
+            initialState
+        )
+    })
+
+    it('should fetch request', () => {
+        const testState: IUiState = {
```
jcaianirh: this is not needed. you are just using testState.loading which you can just pass the boolean value t or f.

src/__tests__/store/ui/reducer.test.ts
```diff
@@ -0,0 +1,180 @@
...
+            isDropdownOpen: false,
+            isKebabDropdownOpen: false,
+            isSidebarOpen: true,
+            sidebarActiveItem: "dashboard",
+            sidebarActiveGroup: "feeds_grp"
+        };
+
+        expect(uiReducer(initialState, {
+            type: UiActionTypes.FETCH_COMPLETE,
+            payload: testState.loading
```
jcaianirh: this is good that you are testing the loading state, you just don't need the const above. just pass in the boolean payload

xhyzzZ:
some errors

PR #21, opened 2019-04-04
update reducer test of message and ui

xhyzzZ:


src/__tests__/store/message/reducer.test.ts
```diff
@@ -0,0 +1,176 @@
+import { messageActionTypes, IMessageState } from "../../../store/message/types";
+import { messageReducer } from "../../../store/message/reducer";
+
+const initialState: IMessageState = {
```
jcaianirh: @xhyzzZ there are currently no reducer modifications on the message state.  you can create a test bed by ensuring there are no mods to the initial state. In this case all tests can be similar to the initial state test. you don't need any testState in this file since there are no modification of the initial state.

src/__tests__/store/ui/reducer.test.ts
```diff
@@ -0,0 +1,141 @@
...
+describe('ui reducer', () => {
+    it('should return the initial state', () => {
+        expect(uiReducer(undefined, {
+            type: null
+        })).toEqual(
+            initialState
+        )
+    })
+
+    it('should fetch request', () => {
```
jcaianirh: the test state is not needed here. As you can see by the reducer there are no inputs to the fetchrequest action. We are just letting the reducer know the fetch request has started and to toggle the loading state. No payload required.

src/__tests__/store/ui/reducer.test.ts
```diff
@@ -0,0 +1,141 @@
...
+                isDropdownOpen: false,
+                isKebabDropdownOpen: false,
+                isSidebarOpen: true,
+                sidebarActiveItem: "dashboard",
+                sidebarActiveGroup: "feeds_grp"
+            }
+        )
+    })
+
+    it('should fetch complete', () => {
```
jcaianirh: @xhyzzZ same as above.

src/__tests__/store/ui/reducer.test.ts
```diff
@@ -0,0 +1,141 @@
...
+                isDropdownOpen: false,
+                isKebabDropdownOpen: false,
+                isSidebarOpen: false,
+                sidebarActiveItem: "dashboard",
+                sidebarActiveGroup: "feeds_grp"
+            }
+        )
+    })
+
+    // it('should set sidebar active item', () => {
```
jcaianirh: @xhyzzZ are you able to make these work?

src/__tests__/store/message/reducer.test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+        )
+    })
+
+    it('should display message', () => {
+        expect(messageReducer(initialState, {
+            type: messageActionTypes.DISPLAY_MESSAGE,
+            payload: testState
+        })).toEqual(
+            {
```
jcaianirh: simplify this to use initialState instead of repeating the contents of the obj every time. this test case and all others.

xhyzzZ: Already deleted

PR #40, opened 2019-07-11
Add Create Feed wizard

### LINKED ISSUE #35: (CHRIS UI - P1 - Development) Create a New Feed
As a ChRIS User, I should be able to create a new feed using the UI.

matankb:
In this PR:
* Added the Create Feed wizard, which allows a user to create a new feed through the UI.
* Updated types file for the `chrisapi` library to work better with the latest version (not all of the new methods are defined, only the ones used in CreateFeed).
* Query individual plugin instances with `status: started` when a feed is loaded. This triggers file registration and updates the status if the plugin is finished (this is part of this PR because it's a step of the feed creation process that was previously done via the command line).

Fixes #35


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@fnndsc/chrisapi": "0.0.4",
+    "@fnndsc/chrisapi": "0.0.5",
```
jcaianirh: would you mind bumping this to 0.0.6 and giving a quick test to make sure everything still works :)

src/__tests__/store/feed/action.test.ts
```diff
@@ -98,4 +101,12 @@ describe("feed actions", () => {
     }
     expect(destroyFeed()).toEqual(expectedAction)
   })
+
```
jcaianirh: Thanks for adding to the tests!

package.json
matankb: Bumped and everything seems fine. I think the bug fix might let us remove the extra `client.getFeeds()` calls - I'll look into that.

src/components/feed/CreateFeed/BasicInformation.tsx
```diff
@@ -0,0 +1,110 @@
+import React from 'react';
+
+import { Form, FormGroup, TextInput, TextArea } from '@patternfly/react-core';
+import Client, { Tag } from '@fnndsc/chrisapi';
+import { Typeahead } from 'react-bootstrap-typeahead';
+
+export declare var process: {
```
jcaianirh: Is this needed here? Is process defined in globals?

src/components/feed/CreateFeed/BasicInformation.tsx
```diff
@@ -0,0 +1,110 @@
...
+      this.setState({
+        availableTagsLoaded: true,
+        availableTags: tags
+      })
+    });
+  }
+
+  async fetchTagList() {
+    const client = new Client(process.env.REACT_APP_CHRIS_UI_URL, { token: this.props.authToken} );
+    await client.getFeeds(); // getFeeds must be called on new Client objects
```
jcaianirh: Hopefully these lines can now be improved as you mentioned in an earlier comment .

src/components/feed/CreateFeed/ChrisFileSelect.tsx
```diff
@@ -0,0 +1,334 @@
...
+
+  render() {
+    
+    const fileList = this.props.files.map(file => (
+      <div className="file-preview" key={file.path}>
+        {
+          file.children ? <FolderCloseIcon /> : <FileIcon />
+        }
+        <span className="file-name">{file.name}</span>
+        <CloseIcon className="file-remove" onClick={() => this.props.handleFileRemove(file) } />
```
jcaianirh: If you can bind this in the constructor and call this on click function without creating an anonymous function it would be more performant.

src/components/feed/CreateFeed/CreateFeed.tsx
```diff
@@ -0,0 +1,507 @@
...
+  closeCreateWizard() {
+    this.setState({ wizardOpen: false });
+  }
+
+  toggleCreateWizard() {
+    if (this.state.wizardOpen) {
+      this.resetState();
+    }
+    this.setState({
+      wizardOpen: !this.state.wizardOpen
```
jcaianirh: Don't reference state in a setState method when using the object parameter since setState is an async call. Use the functional setState and reference prevState: 
setState((prevState, props) => { wizardOpen = !prevState.wizardOpen;});

src/components/feed/CreateFeed/ChrisFileSelect.tsx
```diff
@@ -0,0 +1,334 @@
...
+          onChange={isChecked => this.handleCheckboxChange(isChecked, node)}
+        />
+        {icon}
+        {this.generateFileName(node)}
+      </span>
+    )
+  }
+
+  render() {
+
```
jcaianirh: try destructuring to clean the names of vars as follows:
const { files, handleFileRemove } = this.props;
const { initialTreeLoaded, visibleTree } = this.state; 
and then you can just reference by names instead of this.state. and this.props.
Try doing this everwhere in the code that it makes sense.

matankb:
Great! I'll work on the changes.

src/components/feed/CreateFeed/ChrisFileSelect.tsx
matankb: Since `handleFileRemove` takes the `file` object as a parameter, I think it would need to be bound within the loop like `onClick={this.props.handleFileRemove.bind(this, file)`. Is that ok/better than what it currently is?

src/components/feed/CreateFeed/BasicInformation.tsx
matankb: process is defined in globals, but `process.env.REACT_APP_CHRIS_UI_URL` is typed as `string | undefined` without this declaration. Since Client takes a `string` as the first param, it needs to be typed as a `string`. An alternative could be something like `new Client(process.env.REACT_APP_CHRIS_UI_URL || '', { ... })`. Would that be better?

src/components/feed/CreateFeed/BasicInformation.tsx
matankb: Turns out 0.0.6 didn't fix that, but 0.0.8 did. Those lines (and the equivalent in ChrisFileSelect have been removed by passing the instantiated client as a prop.

jbernal0019:
@matankb 

The `getItems()` and `getItem(id)` helper methods in list resource  objects are not `async` as they do not make any API request. For instance in file  src/components/feed/CreateFeed/BasicInformation.tsx:

`const tags: Tag[] = await tagList.getItems() || [];`

The `await` is not necessary. The data were already fetched from the API with the previous call:

`const tagList = await this.props.client.getTags();`

So I think the `fetchTagList()` function should simply be:

 ```
 async fetchTagList() {
    const tagList = await this.props.client.getTags();
    const tags: Tag[] = tagList.getItems();
    return tags;
  }
```

src/components/feed/CreateFeed/ChrisFileSelect.tsx
jcaianirh: @matankb the way it is looks ok. thanks!

Issue #39, opened 2019-07-10
(CHRIS UI - P1 - DESIGN) Redesign of the feed details layout


mairin:
This was discussed at the UI meeting @BSS in late June. 

There is some confusion about how the data in the blue column next to the graph vs. the data in the header bar and the white content area below the graph relate to each other in the feed details page.

Shania's pipeline design uses the header and white content area below only for pipeline-wide details, and only node-specific content (related to the highlighted node) appear in the blue column next to the graph.

The feed design needs to be cleaned up to follow this pattern because the lower white content area is more mixed up than this. One wishlist item we discussed was an embedded output viewer in the content area!

mairin:
So this was the first redesign attempted a while back. It makes the lower content area solely focused on node details which I think now is the wrong way, but posting here for historical context.

![feed-details_right-panel-revision_1](https://user-images.githubusercontent.com/799683/61051291-ae96bb00-a3b6-11e9-8012-f2c48fbdf833.png)

mairin:
Okay here's a refresh of the refresh design. :-) Some features to look for:

- Rather than listing out all the parameters for individual plugins in the bottom display, there is a dropdown not unlike github's clone button in the selected node area up on the right hand side of the graph that provides a copy/paste-able text readout of the command run to use the plugin in the feed.
- I've added a UI-visible "ID" field to plugin instances in the feed. (I believe the API docs has an ID per plugin instance so this value could be used.) This is to help make it clear which specific plugin's output / attributes you're looking at in case the same plugin+version is used multiple times in the same feed.
- The bottom content area is now both genericized (applies to all nodes but has a dropdown selector for selecting the output of a specific node) and also is basically an output viewer for the nodes above. I would like the output pane here mapped to the selected node in the graph, so if the user selects another node, the output viewer will update to show the output of the newly-selected node.
- The output viewer is expected to behave the same as the current file explorer in the ChRIS UI. So clicking on a filename will popup a preview window with a prompt to view in the full viewer.

![feed-details_right-panel-revision_2](https://user-images.githubusercontent.com/799683/61303220-239e3200-a7b5-11e9-9878-01af8b3d4d91.png)

mairin:
Here's an update based on @jcaianirh 's feedback this morning. Changelog:
[...]

PR #46, opened 2019-07-26
Feed Details Redesign

### LINKED ISSUE #43: ChRIS UI - P1 - DEVEL: redesign of the details layout of feed details.

matankb:
In this PR:
* Redesigned feed details page, following #39 
* A singleton for an instantiated `Client` object. There were multiple spread-out components that all required access to an authenticated `client`, and so they would each have to have duplicate code that would check if the `process.env` constant existed, get passed the `token` as a prop, etc. Instead, there is a single top-level `ChrisAPIClient` class, with a static `.getClient()` method that returns a single `Client` instance.
     * Extracted existing instances of `Client` to the singleton.

Closes #43


src/components/common/textcopypopover/TextCopyPopover.tsx
```diff
@@ -0,0 +1,105 @@
...
+  onMouseDown?: (event: React.MouseEvent) => void,
+  onBlur?: (event: React.FocusEvent) => void,
+  maxWidth?: string,
+}
+
+interface TextCopyPopoverState {
+  copied: boolean
+}
+
+class TextCopyPopover extends React.Component<TextCopyPopoverProps, TextCopyPopoverState> {
```
jcaianirh: @matankb can you use <ClipboardCopy> in patternfly here? I think it would be helpful

src/components/feed/FileBrowser.tsx
```diff
@@ -0,0 +1,284 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import { Breadcrumb, BreadcrumbItem, Split, SplitItem, PopoverPosition } from '@patternfly/react-core';
+import { FileImageIcon, FileCodeIcon, FileAltIcon, FileIcon, FolderCloseIcon, DownloadIcon  } from '@patternfly/react-icons';
+import { Table, TableHeader, TableBody, TableVariant } from '@patternfly/react-table';
```
jcaianirh: awesome! thanks for incorporating the table!

src/components/common/textcopypopover/TextCopyPopover.tsx
matankb: done!

src/components/common/textcopypopover/TextCopyPopover.tsx
```diff
@@ -22,46 +22,12 @@ interface TextCopyPopoverProps {
   maxWidth?: string,
```
jcaianirh: need a comma after className prop (may not be your code :) )

matankb: Fixed

src/components/common/textcopypopover/TextCopyPopover.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+
+import { Popover, TextArea, ClipboardCopy } from '@patternfly/react-core-updated';
+import { CopyIcon } from '@patternfly/react-icons';
```
jcaianirh: unused import for copy icon here

src/components/feed/NodeDetails.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+import TextCopyPopover from "../common/textcopypopover/TextCopyPopover";
+
+interface INodeProps {
+  selected: IPluginItem;
+  descendants: IPluginItem[];
+}
+
+interface INodeState {
+  plugin?: Plugin // the plugin which the currently selected instance is an instance of
```
jcaianirh: missing comma

matankb: Fixed

src/components/common/textcopypopover/TextCopyPopover.tsx
matankb: Fixed

Issue #166, opened 2021-01-19
Background image at /login does not work in Firefox (works in Chrome/Chromium only)


jennydaman:

somya51p:
@jennydaman I wish to work on this issue. I am an Outreachy applicant.

jennydaman:
I found the problem.
https://github.com/FNNDSC/ChRIS_ui/blob/4c8564572be56c254abf0ce399f1d1a850618760/src/assets/scss/components/_background-image.scss#L22

It is a known issue that Firefox does not support `filter` in the same way the Chrome/Chromium does.
Well, this is a strange use anyway. It looks like @PintoGideon was playing with this, given there's a commented out line of code above. Is there a purpose to using `filter` here? We are setting the background using `background-image` anyways.

jennydaman:
@somya51p you may wish to investigate this and consider some refactoring. Background image definition is all over the place

- `assets/scss/components/_background-image.scss`
- `pages/LogIn/login.scss`

My guess is that the purpose of `assets/scss/components/_background-image.scss` is to be generally useful anywhere we want to give a component a background image. In our [design mockups](https://github.com/FNNDSC/cube-design/blob/master/CHRIS-UI/feed-individual.svg), we see that the appbar has a background image.

jennydaman:
Interesting observation! When I opened the link above to the design mockups, Github converts the SVG into PNG and displays that on the webpage. But opening the raw SVG in a new file, Firefox fails to show the background image for the appbar.

somya51p:
@jennydaman I will go through it and try to resolve it as soon as possible. Thanks for the help.

jennydaman:
@somya51p no rush! I think we should pause to think about what the best way to do things would be and try to keep our code clean. After all, part of why this but wasn't straightforward is because it was hard to find `assets/scss/components/_background-image.scss`. Aside from refactors, would you help us write some comments?

It might be a good idea to base your changes off of a fork of https://github.com/FNNDSC/ChRIS_ui/pull/221

somya51p:
Yeah sure @jennydaman.

PR #214, opened 2021-04-04
Background image working for firefox

somya51p:
This PR solves issue numbered #166.
The background image on the login page is visible in both browsers now i.e firefox as well as chrome.

@jennydaman Please review the same.

Screenshot from 2021-04-04 13-06-13

### DIFF
--- a/src/containers/Layout/Header.tsx
+++ b/src/containers/Layout/Header.tsx
@@ -48,7 +48,7 @@ class Header extends React.Component<IHeaderProps> {
         <Badge key={3} style={BadgeStyle}>
           <span>
             Latest update:{" "}
-            <Moment format="DD MMM YYYY @ HH:mm">{`2021-04-07T17:00:10.297464-04:00`}</Moment>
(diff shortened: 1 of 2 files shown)


jennydaman (changes requested):
Seems to have done the trick for 1920x1080p scrren. However, `background-repeat` and responsiveness remain to be fixed.

Those points were implemented on the `.pf-background::after` pseudo-element in [src/assets/scss/components/_background-image.scss](https://github.com/FNNDSC/ChRIS_ui/blob/0a24c132e98bc25c66a919ef9c89ac0db927cfd6/src/assets/scss/components/_background-image.scss#L10-20). @PintoGideon did you have a rationale for using `::after`? I'm guessing here, `::after` might be to work around `.pf-background` element or parent's `margin` and to set `width: 100%; height: 100%` 




# Actual Screenshots

Narrow screen

half screen

Large screen

# Expected Behavior

Half screen

Large screen

PintoGideon:
Yes @jennydaman , you got it right. The `::after` pseudo class helps us set styles on the background image without affecting the markup of the parent element (in our case the `.pf-bacgkround`). It also overrides the margin by setting a `width:100%` and `height:100%`.

Issue #220, opened 2021-04-08
Update background graphic for ChRIS UI login screen


mairin:
This ticket involves updating the background on this screen:
![Screenshot from 2021-04-08 12-03-45](https://user-images.githubusercontent.com/799683/114059573-7f9acb00-9862-11eb-88e4-b26e678bf884.png)

Fixing this issue will include the following tasks:
- [ ] Use optipng or a similar tool to optimize the background graphic (attached below)
- [ ] Create scaled versions of the graphic for the various sizes used by pattern fly. The attached graphic below is the same dimensions as the default largest size.
- [ ] Update the background of the ChRIS UI login screen with the image, such that the correct size of the image loads responsively

![bkgd-chris-nodes](https://user-images.githubusercontent.com/799683/114059468-65f98380-9862-11eb-83b2-4f284a021350.png)

Please feel free to ask any questions / doubts you might have on this task. Cheers!

zrthxn:
@mairin I'd like to work on this if that's alright.

mairin:
@zrthxn Sure thing!

zrthxn:
Just one question, do you think it's necessary in this case to optimize the graphic? Because the total size of all the image files (for all sizes) is currently coming out to be about equal to that of the largest one we had previously (x2000). So the load time gained by optimization would be very small.

mairin:
@zrthxn Yep I do! It doesnt hurt at all and takes like 10-15 min unattended.

PR #216, opened 2021-04-07
Dicomviewer

Zoey-dev:
What does this PR do?

A branch was created and the Dicomviewer component was worked on. The Tools bar was centralised to the middle of the page and the font-size of the icons were reduced to fit properly on smaller screens. Resolves #207 

List of task completed

-Centralised tools bar
-Reduced the font size of the icons

PR #221, opened 2021-04-08
Updated background graphic for ChRIS UI login screen

zrthxn:
Fixes: #220 

- [x] Optimized the background graphic
- [x] Created scaled versions of the graphic for the various sizes used by pattern fly
- [x] Updated the background of the ChRIS UI login screen with the image


zrthxn:
@mairin 
The previous commits are to be ignored. They were my previous work which I had rolled back but I ran into a few merge conflicts while pulling the latest from `master`. Please only consider the last two, i.e. `7ffce9b` and `ddbae03`

mairin:
Let's not worry about the conflict, we can ask that PR be rebased with this if this gets merged first.

Can you please clean up the commits so the PR is clean?

zrthxn:
Okay I'll just do that

mairin (commented):
Hey @zrthxn - I don't think the package-lock.json file should be included in the commit, can you drop it?

I am not sure what happened with the package.json but the diff isn't readable. Can you fix it?

mairin:
@zrthxn It looks like *all* of the diffs are unreadable. What did you use to edit the files?

zrthxn:
The diff of the image files wont be readable since they're binary files. I can see the diffs of the files in the last two commits though. 
I'm just using VS Code so its not because of the editor. Probably because of the changes I pulled and had conflicts.

zrthxn:
I'm just fixing the tree, I know this is a bit of a silly mistake.

zrthxn:
Oh okay. Yes I'll just fix that.
I'll create a new branch, reset it to the master of this repo (not my fork), and rebase the relevant commits onto that from my local master.

zrthxn:
Fixed the problems. Sorry about the mess.

mairin (commented):
Wondering why there is a new sizing introduced here and the 2000 px becoming a new xl?

zrthxn:
The 1200px one wasn't being used anywhere so I assumed that's supposed to be the large one. And when I tried different screen sizes around 900 to 1200px the ChRIS logo ended up in the right area.
If that's wrong I'll change it.

mairin:
> The 1200px one wasn't being used anywhere so I assumed that's supposed to be the large one. And when I tried different screen sizes around 900 to 1200px the ChRIS logo ended up in the right area.
> If that's wrong I'll change it.

The only thing im worried abt is messing with what patternfly does... for example is xl a default in patternfly or defined by patternfly? if not, it could break things in unpredictable ways. If it is a patternfly standard then its fine. Does that make sense?

zrthxn:
Yea I get it. I'll check if that's the case otherwise I'll remove that and
mark this for review.

PintoGideon:
Hey @mairin , I have tested the code for the following devices and it looks good to me. Let me know if this PR is ready to be merged.

**Mobile**:

Mobile


**Kindle**:

Kindle

**Laptop**:

Medium_Screen


**Large Screen**:

Large_Screen

mairin (changes requested):
For Kindle and Mobile sizes.... there's no logo to identify what you are logging into. The ChRIS logo is right now baked into the background. If I provide you the plain ChRIS logo as a separate transparent-background file, would it be possible to pop that logo into the blue space above the login fields?
ChRISlogo-white

zrthxn:
Sure I can do that. If possible you can give me the background separately as well so I can put it in the right place for other screens too. I'll just alter the layout a bit to make it work better.

mairin:
@zrthxn Yay, that's awesome! Here is the background sans logo:
bkgd-chris-nodes_no-logo

jennydaman:
@mairin what's the target min screen size? We should document it somewhere.
https://github.com/FNNDSC/ChRIS_ui/pull/216#discussion_r609676672

mairin:
@jennydaman 360 px is the most common width for mobile phones according to a bunch of stats I just looked up - it's certainly the dimensions of my old samsung galaxy, so we could just go with 360 for the minimum.

The breakpoints for PatternFly are here, I don't see a min there (they have a break at 0?) so I would say if we were to pick one it should be 360, and I am surprised PatternFly doesn't have that anywhere:
https://www.patternfly.org/v4/developer-resources/global-css-variables#global-css-variables

zrthxn:
@mairin 
So I've used the `brandImg` attribute that Patternfly gives you and I've slightly overridden some of the styles to get the following two layouts. I've also modified the logo you provided to make it inline, rather than vertical, in order to fit mobile better.

Desktop
Mobile

I've also added the show-password button for which another issue was created but it was a one-line fix.
If you'd rather have the logo as is, this is what it looks like on a normal screen.

Desktop2

**If these are okay then I can continue fine tuning and create all the scaled versions.**

mairin:
@zrthxn This is looking fabulous, thanks for your attention to detail on this. My only doubt here, is whether we just drop the explanatory text below the logo or replace it. It might be nice to have a short text to just explain that this is the main ChRIS UI... but my gut feel is we just drop it. I just posted a thread in Slack to see what others think - 

https://fnndsc.slack.com/archives/CE6GUG4D6/p1618413787003900

zrthxn:
After changing the background according to the discussion on Slack and adding the appropriate logos and text, this is what the page finally looks like.

Screenshot (49)

I've also gone ahead and added a show password button by using PatternFly's built-in attribute for this purpose.

mairin (approved):
LGTM

PR #229, opened 2021-04-13
Background image visible in Firefox

somya51p:
I have checked, filter option is not changing the image in the background so I have commented on it. Now the image is visible in firefox as well. 
This is a pull request for issue number #166.
@jennydaman @PintoGideon Please review the PR and suggest the needful changes.

Screenshot from 2021-04-13 23-05-57

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "@fnndsc/chrisapi": "^1.5.0",
     "@patternfly/patternfly": "^4.87.3",
-    "@patternfly/react-core": "^4.97.2",
+    "@patternfly/react-core": "^4.106.2",
     "@patternfly/react-icons": "^4.9.2",
     "@patternfly/react-styles": "^4.8.2",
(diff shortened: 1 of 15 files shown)


src/assets/scss/components/_background-image.scss
```diff
@@ -19,7 +19,7 @@
     width: 100%;
     z-index: -1;
     // filter: url("../images/chris_ui_background_filter.svg#image_overlay");
-    filter: url("#{$image-url}#{$chris-background-filter}"); //$chris-background-filter
+    // filter: url("#{$image-url}#{$chris-background-filter}"); //$chris-background-filter
```
jennydaman, line 22: @PintoGideon is this ok?

PintoGideon: Yes @jennydaman , I don't remember why I added the filter. I added it a while ago when I was implementing the dark theme for the UI.

Issue #230, opened 2021-04-13
Show Password functionality in login page


somya51p:
The signup page has the functionality of show passwords, what if the login page also has this functionality. I wish to work on this issue. Please go through it. @jennydaman @PintoGideon

jennydaman:
Have at it.

I would consider using an icon, right-aligned in the box, and maybe show the password only as the button is being held down (hiding on mouseUp).

zrthxn:
@jennydaman Patternfly already has this functionality built-in. Its only an attribute to add in the LoginForm component. It'll probably be add in the fix for #221 in which I'm fixing the login page layout.

PR #232, opened 2021-04-14
Show/Hide functionality added

somya51p:
As patternfly already has this functionality, I have added the same. @jennydaman Please review and suggest changes if any. 
This PR is associated to issue number #230.
Also, I noticed that the signup page is built with functional components, and the login is built with class components. Does it need to be changed? Suggest the same.


jennydaman:
Pretty good! Do you want to port this feature to the sign-up page too?

somya51p:
Yeah sure @jennydaman , I can give it a try.

PR #238, opened 2021-04-15
Show password icon in Signup

somya51p:
I have added the show password icon on the signup page, As someone has already added it to the login page, although issue number #230 was already associated with it.
I have created a new PR(older one #232).  Please review the same and suggest the changes if any @jennydaman @mairin @PintoGideon

### DIFF
--- a/src/pages/SignUp/components/SignupForm.tsx
+++ b/src/pages/SignUp/components/SignupForm.tsx
@@ -15,7 +15,11 @@ import ChrisApiClient from "@fnndsc/chrisapi";
 import { Link } from "react-router-dom";
 import { has } from "lodash";
 import { validate } from "email-validator";
-import {setAuthToken} from '../../../store/user/actions'
+import {setAuthToken} from '../../../store/user/actions';
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faEye } from "@fortawesome/free-solid-svg-icons";
+
+const eye = <FontAwesomeIcon icon={faEye} />;
 
 
 type Validated = {
@@ -152,8 +156,8 @@ const SignUpForm: React.FC<SignUpFormProps> = ({
     }
   };
 
-  const handleShowPassword = (checked: boolean) => {
-    setShowPassword(checked);
+  const handleShowPassword = () => {
+    setShowPassword(showPassword ? false : true);
   };
 
   return (
@@ -220,6 +224,7 @@ const SignUpForm: React.FC<SignUpFormProps> = ({
         validated={passwordState.validated}
       >
         <TextInput
+          style={{width:"93%", position:"relative",}}
           validated={passwordState.validated}
           value={passwordState.password}
           isRequired
@@ -233,17 +238,17 @@ const SignUpForm: React.FC<SignUpFormProps> = ({
               password: value,
             })
           }
-          style = {{
-            marginBottom:"1rem"
+        />
+        <Button
+        style={{
+          position: "fixed",
+          paddingRight: "0.5rem",
+          paddingLeft: "0.5rem",
+          background: "gray" ,
+          borderRadius: "5px"
           }}
-        />
-        <Checkbox
-          isChecked={showPassword}
-          label="Show password"
-          aria-label="Show password"
-          id="showPassword"
-          onChange={handleShowPassword}
-        />
+        onClick={ handleShowPassword }>{eye}
+        </Button> 
        
       </FormGroup>
       
--- a/src/pages/SignUp/components/SignupForm.tsx
+++ b/src/pages/SignUp/components/SignupForm.tsx
@@ -28,7 +28,6 @@ interface SignUpFormProps {
   setAuthToken: (auth: { token: string; username: string }) => void;
   passwordLabel?: string;
   passwordValue?: string;
-  onChangePassword?: (value: string, event: React.FormEvent<HTMLInputElement>) => void;
   isValidPassword?: boolean;
   isShowPasswordEnabled?: boolean;
   showPasswordAriaLabel?: string;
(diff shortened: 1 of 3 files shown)


PintoGideon (changes requested):
Hi @somya51p , The eye icon should be present in the `@patternfly/react-icons library`? Is there a reason we have included the following packages.
```
 "@fortawesome/fontawesome-svg-core": "^1.2.35",
    "@fortawesome/free-solid-svg-icons": "^5.15.3",
    "@fortawesome/react-fontawesome": "^0.1.14",
```

The show password icon in the Login Form is actually a `button`. You can try doing something like this for the SignUp form. 

```
<Button icon={<Eye/>}/>
```
Let me know what you think.

somya51p:
@PintoGideon I was unable to find the eye icon in the @patternfly/react-icons library. So I imported the same from those packages. But now if you are saying I will definitely give it a try.

PintoGideon:
Hey @somya51p , weirdly, the eye icon was not present in the patternfly icon documentation , but I saw that the LoginForm component used it under the hood. 

`import {EyeIcon, EyeSlashIcon} from '@patternfly/react-core';`

Also, if you inspect the code for the LoginForm  (navigate to the <LoginForm/> component in VS code, right click and click on 'Go to Definition') , you can see more logic written that we can use for our `SignupForm` component. 


```javascript

 <FormGroup
        label={passwordLabel}
        isRequired
        validated={isValidPassword ? ValidatedOptions.default : ValidatedOptions.error}
        fieldId="pf-login-password-id"
      >
        {isShowPasswordEnabled && (
          <InputGroup>
            {passwordInput}
            <Button
              variant="control"
              onClick={() => setPasswordHidden(!passwordHidden)}
              aria-label={passwordHidden ? showPasswordAriaLabel : hidePasswordAriaLabel}
            >
              {passwordHidden ? <EyeIcon /> : <EyeSlashIcon />}
            </Button>
          </InputGroup>
        )}
        {!isShowPasswordEnabled && passwordInput}
```

It will help us keep both the forms consistent in terms of styling and functionality.

somya51p:
@PintoGideon I have added that Eye button but don't know why the writing password option is not getting active. Please have a look once and tell me my mistake. I would correct it as soon as possible.

src/pages/SignUp/components/SignupForm.tsx
```diff
@@ -28,12 +26,25 @@ type Validated = {
...
 interface SignUpFormProps {
   setAuthToken: (auth: { token: string; username: string }) => void;
+  passwordLabel?: string;
+  passwordValue?: string;
+  onChangePassword?: (value: string, event: React.FormEvent<HTMLInputElement>) => void;
+  isValidPassword?: boolean;
+  isShowPasswordEnabled?: boolean;
+  showPasswordAriaLabel?: string;
+  hidePasswordAriaLabel?: string;
 };
```
PintoGideon, line 37: The only prop being passed to the `SignUpForm.tsx` is the `setAuthToken` right? This is where the implementation of the `SignupForm` is different to the `LoginForm`. I would probably only use that as a reference.

somya51p: I have passed all required props in SignUpForm.tsx. Then also I need to use only that as reference?

src/pages/SignUp/components/SignupForm.tsx
```diff
@@ -156,9 +167,19 @@ const SignUpForm: React.FC<SignUpFormProps> = ({
...
+
+  const passwordInput = (
+    <TextInput
+      isRequired
+      type={passwordHidden ? 'password' : 'text'}
+      id="pf-login-password-id"
+      name="pf-login-password-id"
+      validated={isValidPassword ? ValidatedOptions.default : ValidatedOptions.error}
+      value={passwordValue}
+      onChange={onChangePassword}
```
PintoGideon, line 180: I don't see an `onChangePassword` event handler. We probably need to  set state of the input right?

src/pages/SignUp/components/SignupForm.tsx
PintoGideon: If you look at the SignUp.tsx , these props are not being passed though.

src/pages/SignUp/components/SignupForm.tsx
somya51p: Ohh, in LoginForm.tsx it was used directly so I got confused. I will have a look at it then.

PintoGideon: No worries. Good catch. Simply put,

The value of the `passwordValue` would be an `' '`.  It's being set to a default value.
Simply put, the fix is to handle the `onChangeEvent` on the `TextInput`. Set the value you receive to the `passwordState` and use that as your value.

somya51p: Okay, thanks for your guidance. I will give it a try

somya51p:
@PintoGideon Added the show/hide password functionality like the LoginForm. Please review the same.

Screenshot from 2021-04-18 12-22-35
Screenshot from 2021-04-18 12-22-48

jennydaman (approved):

PintoGideon (approved):
Looks good to me.

mairin:
Are the buttons on this one backwards? It should be an open eye when the password is hidden, and a slash eye when the password is visible. (The icon/button should describe the action to be taken, not the current state of the field)

somya51p:
Okay, @mairin. I will change according to the action to be taken. Thanks for the review. Do I need to create another PR for the same as this one is being merged?

mairin:
@somya51p that I am not sure of - @PintoGideon what kind of workflow would you prefer for this minor fix post-merge?

PintoGideon:
@mairin  Good catch.  @somya51p - Yes you can create a PR for that minor fix.

PR #245, opened 2021-04-19
Add docker-entrypoint.sh for overwriting CUBE url in build/

### LINKED ISSUE #231: Docker entrypoint to overwrite REACT_APP_CHRIS_UI_URL in docker image
The ChRIS_ui takes a long time to build, so using the build off of DockerHub is faster. However, `REACT_APP_CHRIS_UI_URL=localhost:3000` is compiled into the static `build/` files.

For convenience, we could provide a `docker-entrypoint.sh` as `ENTRYPOINT` which uses `sed` (or the like) to overwrite in the static `build/` dir the backend's URL with something given at runtime. e.g.

```bash
docker run --rm -p 3000:3000 -e REACT_APP_CHRIS_UI_URL=http://localhost:12345/api/v1/ fnndsc/chris_ui:latest
```

jennydaman:
Please try out [docker.io/fnndsc/chris_ui:1.4.1-rc.1](https://hub.docker.com/layers/fnndsc/chris_ui/1.4.1-rc.1/images/sha256-a224a4dab891695975b6e9f6f8ce96c86709b016978744300ef245a7bcc34859?context=explore)

- `docker-entrypoint.sh` script uses `find ... | sed ...` to overwrite CUBE's URL in `build/` if the environment variable `REACT_APP_CHRIS_UI_URL` is given a value at runtime
- Also the script downgrades to an (either random or given) underprivileged user before starting the web server.
- replace [serve](https://www.npmjs.com/package/serve) with [sirv-cli](https://www.npmjs.com/package/sirv-cli)

Close #231

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,5 +27,6 @@ COPY --from=builder /app/build /app
 COPY ./docker-entrypoint.sh /docker-entrypoint.sh
 
 ENTRYPOINT ["/docker-entrypoint.sh"]
-CMD ["sirv", "--quiet", "--etag", "--single", "-H", "0.0.0.0", "-p", "3000"]
+ENV HOST=0.0.0.0 PORT=3000
+CMD ["sirv", "--quiet", "--etag", "--single"]
 EXPOSE 3000
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,57 +1,31 @@
+# ChRIS_ui production mode server
 #
-# Docker file for ChRIS ui production server
-#
-# Build with
-#
-#   docker build -t <name> .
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,57 +1,31 @@
...
 
-WORKDIR /app/
+FROM node:14 as builder
 
+WORKDIR /app
 COPY . .
 
-# Build the app for production
-RUN npm install && npm run build 
+RUN npm install
```
jennydaman, line 16: These `RUN` commands are split up to optimize the builder's layer caching

PR #247, opened 2021-04-19
Fixed the Server Down Issue

### LINKED ISSUE #244: There is no feedback when the server is down
When Trying to Log In or Sign Up while the backend server is down there is no real feedback about what is happening.  
- The Sign Up page screen Actually stays in a Loading state forever. 
- The login page Screen just says invalid Credentials.

mocktarissa:
fix #244 
I added a feedback to the User when the server is down instead of just saying 'invalid Credentials'.


src/pages/LogIn/components/LoginForm.tsx
```diff
@@ -51,6 +54,7 @@ class LoginFormComponent extends React.Component<AllProps, IState> {
       );
     } catch (error) {
       this.setState({
+        errorMessage:(()=>error.message==='Bad server response!'?'Invalid Credentials':error.message)(),
```
jennydaman, line 57: `error.message` is not a stable API. Use other fields of `error` instead, perhaps `error.response`
https://github.com/FNNDSC/fnndsc/blob/f9fda903aa791c484123e9b9ec186a7fd0c3ec3a/js/chrisAPI/src/request.js#L179

yarn.lock
```diff
@@ -0,0 +1,14762 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
jennydaman, line 1: Delete `yarn.lock`. This project is managed by `npm`, not `yarn`

jennydaman (changes requested):
Delete `yarn.lock

src/pages/LogIn/components/LoginForm.tsx
```diff
@@ -51,6 +54,7 @@ class LoginFormComponent extends React.Component<AllProps, IState> {
       );
     } catch (error) {
       this.setState({
+        errorMessage:(()=>error.response?'Invalid Credentials':'There was a problem connecting to the server!')(),
```
jennydaman, line 57: I suppose [`@fnndsc/chrisapi`](https://github.com/FNNDSC/fnndsc/tree/master/js/chrisAPI) itself does not have the best exception objects. Maybe a PR there first would help?

But still you could be doing a `error.response.status === 200` or some other strong assertion.

```json
{
  "columnNumber": 12,​
  "fileName": "http://localhost:3000/static/js/vendors~main.chunk.js",
​  "lineNumber": 14154,
​  "message": "Bad server response!",
  "name": "o",​
  "request": XMLHttpRequest { readyState: 4, timeout: 30000, withCredentials: false, ... },
​  "response": { 
    "data": "Bad server response!",
    "status": 404,
    "statusText": "Not Found",
    ...
  },
  "stack": "a super long string..."
}
```

mocktarissa:
That was the first change I tried to make but when the server is down the response object returns null
This is what @fnndsc/chrisapi returns. 
```
{
    "error": {
        "name": "o",
        "request": {},
        "response": null
    }
}
```
The response object is set to null that's why I wrote it that way.

mocktarissa:
@PintoGideon Can you Take a look to check if everything is in order?

src/pages/SignUp/components/SignupForm.tsx
```diff
@@ -118,6 +123,8 @@ const SignUpForm: React.FC<SignUpFormProps> = ({
           passwordState.password
         );
       } catch (error) {
+
+        console.log({error});
```
jennydaman, line 127: Consider [`console.error`](https://developer.mozilla.org/en-US/docs/Web/API/Console/error)

src/pages/LogIn/components/LoginForm.tsx
```diff
@@ -12,106 +13,106 @@ interface IPropsFromDispatch {
...
-        this.state.usernameValue,
-        this.state.passwordValue
+        usernameValue,
+        passwordValue
       );
     } catch (error) {
-      this.setState({
-        showHelperText: true,
-      });
+      console.log({error});
```
jennydaman, line 50: Seeing https://github.com/FNNDSC/ChRIS_ui/pull/247/commits/961ab97d533959d4cfb0a5e68286229cbdfedc0d, is your choice to not use `console.log`? If so, then delete this one too.

jennydaman (approved):

PR #274, opened 2021-06-29
User Library and PACS Query pages

zrthxn:
https://user-images.githubusercontent.com/35369637/123809619-62c8dd80-d90f-11eb-874f-eb76e6a08540.mp4


mairin:
Hi @zrthxn - this is amazing progress in a quick amount of time, wow! Really excited about this. Here is some UX feedback, hope it is helpful.

### Mobile support

Do you expect this to reflow for mobile or is that not something you've started working on yet?

### Initial search box
Screenshot from 2021-06-29 13-51-11

- We should talk to @rudolphpienaar (if you havent already) to confirm what the default search type should be. I suspect it'd be ID/MRN and not Patient Name, but he would know better.
- I would have the in-field labels match the dropdown labels. Right now you have (excepting the hints which are fine as-is):
  - Name => Patient Name   >>> I would recommend both be labeled "Patient Name"
  - Date => Study Date    >>> I would recommend both be labeled "Study Date"
  - ID or MRN => Patient ID or MRN    >>> I would recommend both be labeled "Patient ID or MRN"
- It's my understanding PACS don't allow booleans or partial search queries (eg with * or ?), making Patient Name a much less useful field since the spelling has to be exact to match anything. Is something implemented in-between here to allow that?
- I am concerned that the three search modes are a bit hidden behind the dropdown, but we can do some usability testing to figure out if thast is the case in practice and see if users actually use the dropdown and discover the additional modes. Visually this is a more elegant solution.

### Filters & Constraints label text
#
Screenshot from 2021-06-29 10-39-59

- I think the label language is a bit technical / overly specific and could cause confusion or maybe be a little intimidating? I would just replace the "Filters & Constraints" label with "More lookup options"
- It seems the search mode (name, id/mrn, date) will effect these fields. Have you accounted for that? E.g. if I am searching by study date, having a date field here doesn't make sense so it should be at least greyed out maybe if not display: none. 
- Looking at the next screenshot below, for a search by patient name, the date field in filters & constraints is gone - do you know why?

### Patient results

Screenshot from 2021-06-29 13-52-36

- Gender is highlighted with the tick background a lot here, but I am not sure it is so important to highlight it this far up the visual hierarchy. I would think if you have for example multiple patients here with the same name, the key thing you'd want to notice high up the visual hierarchy to differentiate between them would be DOB. Can you replace the gender notation with DOB XX/XX/XX in the little tick?
- Just a wacky idea that might be helpful for users.... if a study has happened in the past month let's say, where it has the latest study timestamp in the lower right, can we highlight that in a diff color (maybe red, maybe something else) with an icon or "*" and call out something like Last 30 days so the most recent studies jump out? (Or maybe make this additional label simply the human-readable "since" label in bold... eg. "10 days ago")
- I am assuming these results are in reverse chrono based on the date of the most recent study for each patient, is that correct?

### Study results
Screenshot from 2021-06-29 13-52-56

This is a great display! 

- I would suggest listing the station name as a separate column. Just a gut instinct here but the station name indicates *where* the scans were taken and that might be an important clue to tracking down the right study or differentiating between two similar ones.
- The blue ticks for modality feel too visually heavy. Can we do a dark grey instead?
- I would ask @rudolphpienaar if the # of files is important here. I think number of series is good but not sure if # of files is extra noise not needed at this layer? (could be indicated one layer down)
- The date per study feels hard to read - it kind of blends in but seems like it should be higher up the visual hierarchy. Maybe pull in a date/cal icon to call it out a little better? Or maybe just bold it. Not sure.
- We should test this with users but I am worried the interior scroll under each patient record to view studies won't be obvious. Does it work without a scroll wheel - could be an accessibility issue if not. Is there a visible scrollbar somewhere or one that could be turned on?

### Study history stack

Screenshot from 2021-06-29 13-54-40

This is such a cool solution, to have a history stack of viewed studies. This is a really novel solution to a tricky interaction, I love it. A few tweaks:

- "Studies" is a little different than the individual studies in the list, it's sort of like a home button or a menu. I would call it out differently. Can you give it a blue background with white text, have it sort of right up against the vertical bar as a kind of "flag" so it's visually different than the other items in the vert menu and so the eye is drawn more to it so it's clear how to get "back"?
- I would have a little "X" to the right of each study in the history stack, in case someone clicked on something by accident and wanted to clear the stack to help organize their view here.
- I would cut down the instructional text on the top of the list of series per study, and have a little header there with metadata about the selected study instead.
- Consider adding a "< back to list of studies" button that does the same as the "Studies" button at the top of the history stack. Novel interfaces have the issue that ppl might not get it right away so this would be a crutch for folks not used to it / who feel trapped or lost.

### Success

Screenshot from 2021-06-29 13-53-58

Rather than having repetitious "success" messages, I'd leave this blank and only have a label to indicate the failure condition. 

### Series selection


Screenshot from 2021-06-29 13-54-23

Screenshot from 2021-06-29 13-53-35

I love this! This is a nice system to enable multiple series selection. I worry this mechanism won't work on mobile though. Do you have thoughts on that?

zrthxn:
@mairin 

### Mobile support

I have tried to optimize a few things for mobile but its tough to accommodate that browsing layout on mobile. I think we will have to make a judgement call on that based on how prevalent mobile usage is on the platform.

### Initial search box

I'll fix the dropdown labels. They were shortened to save space. 

Yes the name field has to be an exact match which I think is still better than not having it. And I dont think we can implement much in the way of fixing this since if the source of the data itself doesnt allow partial text search it then we're stuck with it.

### Filters & Constraints label text

>It seems the search mode (name, id/mrn, date) will effect these fields.

Not yet but we can. These filters are supposed to be frontend filters just to make finding results easier. So yea they can change as we like.

> Looking at the next screenshot below, for a search by patient name, the date field in filters & constraints is gone - do you know why?

I actually updated the video after creating the issue. Maybe you were still looking at the old version.

### Patient results

Can replace gender yes. 

Yes can add a "NEW" badge in blue.

> I am assuming these results are in reverse chrono based on the date of the most recent study for each patient, is that correct?

Not as of now because we're using some dummy data at the moment. But yes that will be in the final version.

### Study results.

> I would suggest listing the station name as a separate column. 

Okay, that's a good idea. And the accession number isnt meaningful to the user anyway.

> The blue ticks for modality feel too visually heavy. Can we do a dark grey instead?

Definitely. 

When i showed this, the number of files was just there as extra info. The date can be made more visible, although an icon would be too small. 

The scroll does work without a wheel using arrowkeys but yes it is not immediately obvious especially since the boundary of the card lines up such that it looks like the end of the list. It can be made to fit the height of the content but then that'd mean if there are lots of studies, the page becomes huge.

> Is there a visible scrollbar somewhere or one that could be turned on?

They've been hidden explicitly so yes it can be made visible. Although you then have to have both vertical and horizontal bars, not just one.

### Study history stack

will do.

### Success

Agreed. Will fix.

### Series selection

It can be made to fit on different places on the screen, or we can made a custom component to suit this, rather than using the canned PF thing.

zrthxn:
### Incorporated design feedback
Screenshot from 2021-06-30 14-09-30
Screenshot from 2021-06-30 14-09-20

mairin:
Hi @zrthxn ! Great work! Some more points:

- **Filters label:** I still think **"More lookup options"** is the correct label here. The fact that it is a filter or not is more of an implementation detail, the mental model users have for these types of search interfaces for filters is really extra "options" - most commonly called "advanced search" and the term "filter" introduces complication. I am suggesting "More lookup options" because "more options" isn't clear as to what the option is for, and using the term "lookup" creates parallelism with the page header "PACS Lookup System" Does that make sense
- **NEW tick label** What are the criteria for the NEW tick - is it study is within the past 30 days - would it be possible to put a tooltip over it to state that. They might need a little more left side padding from the name of the item.
- **Treatment of Studies link in history stack** - Apologies if this is a WIP and not ready for comment yet but I think this needs to be treated visually still as per earlier comments
- **Series detail** - the study metadata header looks great! The "x" mark - what does that do - does it go back to the studies tab. I suggested adding x to the right of each series' history stack label to remove it from the stack.
- **Modalities tick** - nice pick of grey color, they seem to fit much more nicely now!

zrthxn:
- filters label will be done
- will add tooltip and padding
- still working on the studies link in stack, just a different background isn't looking that good to me. 
- the X mark removes the study from the list and takes you back to the studies list. It isn't possible to add a button inside the tab/link so I added it there.

zrthxn:
@PintoGideon If the feedback is positive then this can be merged but not deployed.

PintoGideon:
Hey @zrthxn ,

Overall, The PR looks quite good and should be ready to merge. A few quick observations that I need your take on.

1. I see a lot of console warnings for invalid dom nesting. Could you perhaps take a look at that? I have attached screenshots 
    where I saw them. I may have to perhaps check my own code for these warnings under Strict mode and fix them.
  Console_Error
console_error_2
console_error_3

2. When the query builder returns a single patient, it should be 1 patient matched your search.
    Similarly, if you have a single study, it should say 1 study.
    
SinglePatient
Single_Study

zrthxn:
@PintoGideon 

1. I'll try to fix most of those. Some I think are because of using certain PF components inside each other.

2. Yea I thought about that too, initially I had written a tiny function to do that but I thought thats a needless complication. I'll add that again.

zrthxn:
### PACS QR
Demo of browser
Screenshot from 2021-07-21 10-24-44

### File explorer
Screenshot from 2021-07-13 21-30-06

zrthxn:
My `.env.local` file if anyone wants to test. 
```
 BROWSER="google-chrome"
 
 # CUBE
 REACT_APP_CHRIS_UI_URL="https://cube.outreachy.chrisproject.org/api/v1/"
 # REACT_APP_CHRIS_UI_URL="https://cube.pangea.chrisproject.org/api/v1/"
 # REACT_APP_CHRIS_UI_URL="http://localhost:8000/api/v1/"
 REACT_APP_CHRIS_UI_USERS_URL="${REACT_APP_CHRIS_UI_URL}users/"
 REACT_APP_CHRIS_UI_AUTH_URL="${REACT_APP_CHRIS_UI_URL}auth-token/"
 
 # Set PFDCM_URL to the root url of the running pfdcm instance
 # REACT_APP_PFDCM_URL="https://pfdcm.outreachy.chrisproject.org/"
 REACT_APP_PFDCM_URL="https://pfdcm.pangea.chrisproject.org/"
 # REACT_APP_PFDCM_URL="http://172.17.0.1:4005/"
```

PR #277, opened 2021-07-01
Add error handling to ChRIS file select

### LINKED ISSUE #273: Prevent app crashes on ChRIS API error
The application crashes while browsing the file tree in the feed creation wizard. Theres no problem in the app itself but if the API returns anything other than status 200, or throws an error the app will crash.

>  Error while fetching feed files o: Internal server error

Screenshot from 2021-06-29 12-34-51

### Expected Behavior 
The error should be handled and some sort of notification shown.

matankb:
This PR adds error handling to the ChRIS file select page in the Create Feed Wizard, and displays an error message if the ChRIS api returns with a non-200 status (such as an internal server error). Previously, the entire app would crash. 

Fixes: #273 

image


PintoGideon (approved):
LGTM.

PR #296, opened 2021-09-09
[Fix] Issues with PACS query UI

zrthxn:
- Fixing ui update issue


zrthxn:
ezgif com-gif-maker

zrthxn:
Things work fine right now. The frontend code is nearly perfect.
However there is a slight problem with the `pfdcm` backend. With long running tasks, like running a "retrieve" or "push" on hundreds of files, the returned status from the API does not match the actual state of the application. This means that, for example, if a "retrieve" thread is started, the API will not respond with an appropriate status until some time later. Similarly if a "push" thread is started, the API will keep responding with a "registering" status.

Issue #287, opened 2021-07-26
Opening any page/link in a new tab requires a new login


zrthxn:
After logging in, if one opens a link from the sidebar in a new tab (with "Open in New Tab") the app doesn't login automatically, as it should. Instead you have to re-login on the new tab

### Expected Behaviour
The app should login automatically, by remembering the auth token, saving it to local storage or as a cookie.

PR #302, opened 2021-09-29
Stability Improvements

zrthxn:
- No infinite spinner on invalid PACSservice
- Rounded cards
- Sidebar text change to "PACS"
- Make filters actually functional
- Fix infinite spinning in library on login

Fixes: #287


zrthxn:
Fixes: #287

PR #313, opened 2021-10-17
Implemented scale nodes via output size

ppk-1996:
I enabled the output size option in the select component and write overlayScale using number_of_workers assuming it is the size of the node. If I did something wrong, please correct me and I'll fix asap.


mairin:
Hi @ppk-1996 - any chance you could provide screenshots and/or a screencast demo of the change?

ppk-1996:
@mairin I think the backend developers are still working on output size issue. I'll continue working on this when they got it figured out.

PR #314, opened 2021-10-17
Lined up quadrants on the feed details screen

ppk-1996:
It will be the best if the bottom panel is also a draggable Drawer, but for now the two top and bottom panels are aligned with changing 50% to 49.12% in top panel size.


ppk-1996:
Thank you for the suggestion. I provided the screenshot that shows the middle bars are line-up appropriately.

PR #339, opened 2021-11-17
Feature/catalog lookup

bolajahmad:


bolajahmad:
This PR solves a couple of UI issues detected while using the CHRIS_UI.

- The items displayed in the catalog is unresponsive on desktop and on tablet. This PR optimizes the space and makes the content more readable
- When an item at the bottom of the page is selected from the catalog, the drawer comes up but unless you know to scroll up, the content is missing. This PR implements scroll to top
- The feed details is also unresponsive

PR #380, opened 2022-04-05
No need to change year date manually in Copyrights

AryaArunB:
There is no need to change the year on the login page every time
Screenshot 2022-04-05 at 8 08 34 PM
.


PintoGideon:
@AryaArunB  Why was the version string removed from the header page and added to the Overview Page?

AryaArunB:
@PintoGideon  Actually the version and latest date spans were not responsive when we had it in the header. Thus I thought that it was better to put it in the Welcome Chris section. imo the header feels cleaner. What do you think?

PintoGideon:
Got you @AryaArunB . It does look cleaner however we would want to keep the version string in the header for maximum visibility. There are times when multiple UI tabs are open which are instantiated from either a local fork or from the latest master build , and glancing over the version string helps us know if the UI is dated or no.

AryaArunB:
@PintoGideon got it. I have retained the version number on the header. Should I make some other changes?

PintoGideon:
@AryaArunB  All good :)

PR #386, opened 2022-04-07
Fixed #385 Broken "Delete a pipeline" Button

### LINKED ISSUE #385: Broken Button "Delete a Pipeline" in the Analysis Catalog Page
**Steps to Reproduce**

This Bug is Present in the Pipelines section of the "Analysis Catalog" page ,
**You will need have pipelines uploaded to see this bug , If you are running minChris as a backend then you'll not see anything in the Pipelines section , because miniChris doesnot has pipelines preuploaded , to upload pipelines refer this --** https://github.com/FNNDSC/caw#testing

After you have your pipelines uploaded go to the "Pipelines" section of "Analysis Catalog" page and **select any  pipeline** , **you'll see a delete pipeline button** , **now on reducing the width of the screen the button will break** , **also you can use chrome Devtools to open the mobile version of the website to see this bug.** I am attaching the screenshots of both take a look.

bug analysis catalog page

Prakashh21:
This fixes #385 

Increase the width of the button slightly and added minWidth.
I have added screenshots below take a look.

button fixed

### DIFF
--- a/src/components/catalog/DisplayPage.tsx
+++ b/src/components/catalog/DisplayPage.tsx
@@ -335,7 +335,8 @@ const DisplayPage = ({
             {showPipelineButton && (
               <Button
                 style={{
-                  width: "45%",
+                  width: "55%",
+                  minWidth: "45%",
                 }}
                 onClick={async () => {
                   if (selectedResource.data.locked === true) {
--- a/src/components/catalog/DisplayPage.tsx
+++ b/src/components/catalog/DisplayPage.tsx
@@ -335,8 +335,8 @@ const DisplayPage = ({
             {showPipelineButton && (
               <Button
                 style={{
-                  width: "55%",
-                  minWidth: "45%",
+                  width: "fit-content",
+
                 }}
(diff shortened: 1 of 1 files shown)


src/components/catalog/DisplayPage.tsx
```diff
@@ -335,7 +335,8 @@ const DisplayPage = ({
             {showPipelineButton && (
               <Button
                 style={{
-                  width: "45%",
+                  width: "55%",
```
jennydaman, line 338: This is a large and dynamic component, so setting width based on some percentage of its parent's size could be a brittle solution.

I think it makes more sense here to have the minimum width be dependent on the size of its contents instead of the size of the parent.

Prakashh21: Alright @jennydaman width:fit-content would do it.. Thanks !! for the feedback .. I have pushed the changes.

jennydaman: It always seems like CSS has these features that you gotta dig around for.

PR #388, opened 2022-04-07
Button for Already have an account

AryaArunB:
Before:
Screenshot 2022-04-07 at 11 12 21 PM
After:
Screenshot 2022-04-07 at 11 12 35 PM


AryaArunB:
@PintoGideon is it better?

Issue #412, opened 2022-04-18
Responsiveness of Dashboard Cards.


aparna-bhatt:
The images been displayed on the dashboard are not responsive even in the full desktop site.
### SCREENSHOT
![Screenshot from 2022-04-18 23-29-22](https://user-images.githubusercontent.com/86338542/163852864-5ffe88a2-091c-4b95-983e-8e99ac5c1dcb.png)
The third image get hidden.

Akanksha2403:
Hey @aparna-bhatt  I would like to work on this issue. can you assign it to me?

aparna-bhatt:
Hey @Akanksha2403 I am a fellow contributor and I have opened this issue and solved it.You can solve some other issues.

aparna-bhatt:
@jennydaman sir can you review my PR #413

AryaArunB:
Hey @aparna-bhatt I think making a carousel/ collage of these pictures might help in making this responsive. I have tried to make a carousel for the same #391

aparna-bhatt:
Yeah I think carousel is looking fine .

PR #391, opened 2022-04-09
Added a carousel slider for the pictures in the overview tab.

aryabyte21:
@PintoGideon, please review it.

PR #413, opened 2022-04-18
Dashboard cards not responsive even in desktop view

aparna-bhatt:
Resolved issue #412 

### BEFORE
Screenshot from 2022-04-18 23-29-22

### AFTER

Screenshot from 2022-04-18 23-28-13

### DIFF
--- a/src/pages/Dashboard/Dashboard.scss
+++ b/src/pages/Dashboard/Dashboard.scss
@@ -10,3 +10,8 @@
     display: none;
   }
 }
+.image-card{
+  display: "flex";
+  flexdirection:"row";
+  flexwrap:"wrap";
+}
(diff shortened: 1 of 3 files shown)


aparna-bhatt:
@jennydaman sir can you review my PR.

jennydaman:
Hmm, we should think about the design here. These images are not necessary at all, they're just examples. It's nice to show many of them when the screen is wide enough, but on small screens, we don't need to show more than one, and I think line wrapping is wasting space here. Do you have any other ideas for how to make this view responsive?

aparna-bhatt:
I am still thinking more on a way to display more images efficiently. I closed this PR as multiple PR's of same kind were created so I will help to contribute in #391.

PR #508, opened 2022-06-29
Self-updating build version string

jennydaman:
"Built on" timestamp replaced by a **self-updating** "build version" string. Its format is as follows:

```
YYYYMMDD.X+commit(-dirty?)

#          YYYYMMDD = current date
#                 X = number of merge commits since the tag version-0
#            commit = HEAD commit short sha
#            -drity = suffix indicating there are uncommitted changes
```

image

The same string (but with illegal characters replaced) is used to tag docker images.

In this proof-of-concept the "build version string" is very verbose. It's ok to merge as-is. Alternatively, we can discuss how we want to change it. It's verbose to show off what's possible. Changes can be made to `print_version.sh`.

In a follow-up PR we can discuss and implement different styles to be applied to the version string (for example, maybe the commit SHA should be dimmer, or the version number in bold?).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,12 +1,3 @@
-.dockerignore
-Dockerfile*
-.gitignore
-.travis.yml
-.git
-LICENSE
-README.md
-
-
 # dev container stuff that might end up here bc of the dev environment
 node_modules
 .config/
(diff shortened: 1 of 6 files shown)


rudolphpienaar:
This seems all acceptable to me.

PR #510, opened 2022-06-30
Tests

Mfonn:
# Background Information
Tests are setup to ensure the code runs as it is supposed to and new changes do not affect the current workflow. 

# Motivation
This PR contains setup for the testing framework, Cypress, tests for the signup, login and Create New Analysis pages.
Screenshot from 2022-07-02 08-35-49
 from 2022-07-02 08-36-22](https://user-images.githubusercontent.com/67243684/176991573-8704052d-a9c5-4462-8b0e-fa12d52534a2.png)
e3-1d6a-4229-ae6e-92d0ad8924a1.png)

# How to Test
- Make sure you have ChRIS backend running on `http://localhost:8000/api/v1/`
- ChRIS frontend should be running on `http://localhost:3000/`
- Install cypress using `npm install cypress --save-dev`
- Run Cypress using  `npm run cypress:open`
- This command opens up the Cypress UI for interactive testing.

### DIFF
--- a/src/components/catalog/ComputeCatalog.tsx
+++ b/src/components/catalog/ComputeCatalog.tsx
@@ -6,7 +6,7 @@ const ComputeCatalog = () => {
   const [computeResources, setComputeResources] = React.useState<any[]>();
   const [pageState, setPageState] = React.useState({
     page: 1,
-    perPage: 5,
+    perPage: 10,
     search: "",
     itemCount: 0,
   });
--- a/src/components/catalog/PipelineCatalog.tsx
+++ b/src/components/catalog/PipelineCatalog.tsx
@@ -1,87 +1,88 @@
-import React, { useEffect } from "react";
-import ChrisAPIClient from "../../api/chrisapiclient";
-import DisplayPage from "./DisplayPage";
+import React, { useEffect } from 'react'
+import ChrisAPIClient from '../../api/chrisapiclient'
+import DisplayPage from './DisplayPage'
 
 const PipelineCatalog = () => {
-  const [pipelines, setPipelines] = React.useState<any[]>();
-  const [fetch, setFetch] = React.useState(false);
-  const [filteredId, setFilteredId] = React.useState<number>();
+  const [pipelines, setPipelines] = React.useState<any[]>()
+  const [fetch, setFetch] = React.useState(false)
+  const [filteredId, setFilteredId] = React.useState<number>()
   const [pageState, setPageState] = React.useState({
     page: 1,
-    perPage: 5,
-    search: "",
+    perPage: 10,
+    search: '',
     itemCount: 0,
-  });
+  })
 
-  const { page, perPage, search } = pageState;
-  const [selectedPipeline, setSelectedPipeline] = React.useState<any>();
+  const { page, perPage, search } = pageState
+  const [selectedPipeline, setSelectedPipeline] = React.useState<any>()
 
   const onSetPage = (_event: any, page: number) => {
     setPageState({
       ...pageState,
       page,
-    });
-  };
+    })
+  }
   const onPerPageSelect = (_event: any, perPage: number) => {
     setPageState({
       ...pageState,
       perPage,
-    });
-  };
+    })
+  }
 
   const handleFilterChange = (value: string) => {
     setPageState({
       ...pageState,
       search: value,
-    });
-  };
+    })
+  }
   useEffect(() => {
     async function fetchPipelines(
       perPage: number,
       page: number,
-      search: string
+      search: string,
     ) {
-      const offset = perPage * (page - 1);
-      const client = ChrisAPIClient.getClient();
+      const offset = perPage * (page - 1)
+      const client = ChrisAPIClient.getClient()
       const params = {
         limit: perPage,
         offset: offset,
         name: search,
-      };
-      const pipelinesList = await client.getPipelines(params);
+      }
+      const pipelinesList = await client.getPipelines(params)
 
-      let pipelines = pipelinesList.getItems();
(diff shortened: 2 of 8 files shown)


cypress/e2e/CreateNewAnalysis.cy.ts
```diff
@@ -0,0 +1,86 @@
...
+  
+    })
+
+    it('Clicks on Create New Analysis', () => {
+        cy.get('.create-feed-button')
+        .click()
+    })
+
+    it('Types in the Feed Name', () => {
+        cy.get('input#feed-name').type('MRI Study')
```
jennydaman, line 14: This comment does not need to be addressed for this PR but please think about them 

Consider having the string "MRI Study" include a random component. Then write an _assertion_ to check whether or not the feed title was actually set correctly by the ChRIS_ui.

For example, I am writing some tests for a project I am working on, [chrs](https://github.com/FNNDSC/chrs). Its automatic tests create pipelines with random names and random descriptions. These are just two random words.

screenshot

jennydaman (approved):

jennydaman:
Good work. @PintoGideon merge this if you are okay with it.

Next we need to work on automatically running these tests using Github Actions.

cypress/e2e/02CreateNewAnalysis.cy.ts
```diff
@@ -0,0 +1,106 @@
...
+    
+    it('04 Goes to FeedType selection', () => {
+        cy.get('button.pf-c-button.pf-m-primary')
+        .eq(1)
+        .click()
+      })
+
+    it('05 Selects the feedtype', () => {
+        cy.get('label.pf-c-radio__label')
+        .eq(0)
```
jennydaman, line 34: Somehow make sure this is actually the button you want, either via assertion or assign the element a test ID.

jennydaman (changes requested):

cypress/e2e/02CreateNewAnalyses.cy.ts
```diff
@@ -0,0 +1,124 @@
+///<reference types="cypress" />
+
+describe('Sign In user', () => {
+  it('logs in as chris user', () => {
+    cy.visit('http://localhost:3000/login')
+    cy.get("input#pf-login-username-id").type("chris")
```
jennydaman, line 6: This is ok for now, but it would be better to run your tests as the newly created fake user instead of the "chris" user.

It's best to use randomized data whereever possible instead of "default" or "well-known" values because programmers may (either accidentally or intentionally) hard-code behavior dependent on those values. This is true for the ChRIS backend: the "chris" user is hard-coded to have special permissions which no other user has, and so if you test using the "chris" user, your tests might show results which wouldn't be valid for any other user.

cypress/e2e/02CreateNewAnalyses.cy.ts
```diff
@@ -0,0 +1,124 @@
...
+        cy.get('button.pf-c-button.pf-m-primary')
+        .eq(1)
+        .should('have.text', 'Close')
+
+        .click()
+    })
+
+    it('15 Clicks on the newly created analysis', () => {
+      cy.get('span.feed-list__name')
+      .eq(0)
```
jennydaman, line 118: Are you sure that the 0th item in this list is the feed you just created? You should make an assertion that the feed name is correct.

For that reason, you should also be using `faker` to randomly create a feed name as well, or you could use a timestamp in the feed name, so that this current test run looks different from past test runs.

package.json
```diff
@@ -28,6 +28,7 @@
     "d3-zoom": "^1.8.3",
     "dicom-parser": "^1.8.11",
     "email-validator": "^2.0.4",
+    "faker": "^5.5.3",
```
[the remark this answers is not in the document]
jennydaman: :bangbang: Please address this @Mfonn

package.json
```diff
@@ -65,6 +66,8 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "lint": "eslint '*/**/*.{js,ts,tsx}' --quiet --fix",
+    "cypress:open": "cypress open",
+    "cypress:run": "npx cypress run",
```
jennydaman, line 70: :bangbang: Delete `npx`, it is not necessary here.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,32 @@
+# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: E2E on Chrome
```
jennydaman, line 4: :bangbang: Change the name to simply "Tests" (we're not specifically using Chrome here)

README.md
```diff
@@ -144,3 +144,25 @@ To learn React, check out the
...
 [repo-link]: https://github.com/FNNDSC/ChRIS_ui
 [code-size]: https://img.shields.io/github/languages/code-size/FNNDSC/ChRIS_ui
+
+
+### E2E TESTS ARE RAN USING CYPRESS
+
+## Prerequisites:
+- ChRIS_ultron_backend is running on `http://localhost:8000/api/v1/`
+- ChRIS_ui is running on `http://localhost:3000/`
+- You have Cypress installed using `npm install cypress`
```
jennydaman, line 154: :bangbang:  `npm install cypress` is the wrong command, you simply need `npm install` which reads and installs the correct versoin from `package.json`

jennydaman (changes requested):
I have left two kinds of comments:

- comments starting with :bangbang: must be resolved before merging.
- other comments can be deferred as future work

PR #657, opened 2022-10-11
fixed date picker ui error  in /pacs

mightyFZeus:


mightyFZeus:
this is was the initial ui bug on the date pick
Screenshot (94)


now it looks like this

Screenshot (95)



Screenshot (96)

mightyFZeus:
@jennydaman  @PintoGideon  @mairin  as assigned to me, I have corrected the date picker UI bug in /pacs. Please review...Thanks

jennydaman:
@mightyFZeus thank you for the screenshots. Are you able to figure out why `<DatePicker>` does not look as advertised? We would prefer to use the Patternfly elements where possible,

https://www.patternfly.org/v4/components/date-picker/react-demos

mightyFZeus:
@jennydaman thanks for the reply, I will look into it and make necessary adjustments

mightyFZeus:
@jennydaman  i have made necessary adjustments... now the patternfly/react- core datepicker componets is used


Screenshot (97)
Screenshot (98)

jennydaman (approved):

jennydaman:
Wonderful! I think it's always worth it to dig a little deeper into the problem.

PR #681, opened 2022-10-13
Copy-Pastable Chrs download command

### LINKED ISSUE #404: Copy-pastable chrs download command
[chrs](https://github.com/FNNDSC/chrs) is a command-line client for _ChRIS_. In some cases it is advantageous to use a CLI tool, one example would be for downloading a large number of files. The command [`chrs download`](https://github.com/FNNDSC/chrs/tree/master/chrs#chrs-download-src-dst) is an efficient way to download a large number of files from _ChRIS_. This is a feature request for the "My LIbrary" screen of *ChRIS_ui* to have a "copy chrs download command" which copies into the clipboard a command to download the current path being shown in the UI.

For reference, here are some other UI projects with similar features:

### IPFS

image

### OpenShift

image

ciradu2204:
**Description:** 
We added a function to allow users to copy the chrs download command based #404 issue. The command `chrs download`  is an efficient way to download a large number of files from ChRIS. 

Screen Shot 2022-10-26 at 15 39 55

**How to test the feature**
Prerequisite: Download the [chrs](https://github.com/FNNDSC/chrs) cli. 
1. click the copy command button as seen in the above image
2. paste the command in the chrs cli to download the files of that specific feed

**Result** 

Screen Shot 2022-10-15 at 20 35 35

fixes #404

### DIFF
--- a/src/components/feed/NodeDetails/NodeDetails.tsx
+++ b/src/components/feed/NodeDetails/NodeDetails.tsx
@@ -1,4 +1,5 @@
 import React, { Fragment, useState } from "react";
+
 import Moment from "react-moment";
 // Added Tooltip
 import {
@@ -68,7 +69,7 @@ const NodeDetails: React.FC<INodeProps> = ({ expandDrawer }) => {
   
   // Method 2.
   const { id } = useParams();
-  const baseUrl = window.location.origin;
+  const baseUrl = process.env.CHRIS_UI_URL;
   const copyText = "Copy To Clipboard";
   const doneCopyText = "Copied!"
   const [isCopied, SetIsCopied] = React.useState(false);
@@ -113,7 +114,7 @@ const NodeDetails: React.FC<INodeProps> = ({ expandDrawer }) => {
 
   // Method 2
   const copyPathToClipboard = () => {
-    const path = `${baseUrl}/api/v1/plugins/instances/${id}/files/`
+    const path = `${baseUrl}/plugins/instances/${id}/files/`
     const command = `chrs download ${path}`
     navigator.clipboard.writeText(command);
     SetIsCopied(true);
--- a/src/components/feed/NodeDetails/NodeDetails.tsx
+++ b/src/components/feed/NodeDetails/NodeDetails.tsx
@@ -69,7 +69,7 @@ const NodeDetails: React.FC<INodeProps> = ({ expandDrawer }) => {
   
   // Method 2.
   const { id } = useParams();
-  const baseUrl = process.env.CHRIS_UI_URL;
+  const baseUrl = process.env.REACT_APP_CHRIS_UI_URL;
   const copyText = "Copy To Clipboard";
   const doneCopyText = "Copied!"
   const [isCopied, SetIsCopied] = React.useState(false);
--- a/src/components/feed/NodeDetails/NodeDetails.tsx
+++ b/src/components/feed/NodeDetails/NodeDetails.tsx
@@ -244,10 +244,7 @@ const NodeDetails: React.FC<INodeProps> = ({ expandDrawer }) => {
             </Button>
 
             {/* Editing button */}
-            
-            {/* <div style={{ margin: '100px' }}> */}
-            {/* const PopoverBasic: React.FunctionComponent = () => ( */}
-              <div style={{ margin: '50px' }}>
+              <div>
                 <Popover_patternfly
                   aria-label="Basic popover"
                   headerContent={<div>Popover header</div>}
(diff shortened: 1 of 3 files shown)


src/components/feed/NodeDetails/NodeDetails.tsx
```diff
@@ -95,6 +108,14 @@ const NodeDetails: React.FC<INodeProps> = ({ expandDrawer }) => {
     pluginParameters,
   ]);
 
+  const path = `${baseUrl}plugins/instances/${id}/files/`
```
jennydaman, line 111: String concatenation does work, but it's better to rely on the API to tell us what the correct link is instead of hard-coding it.

I'm more familiar with the backend Collection+JSON API than I am with the front-end code. Knowing that the response on `/api/v1/plugins/instances/N/` is going to return a link to `/api/v1/plugins/instances/N/files/`, the value for `path` that you want already exists as a variable in some object somewhere, and you just need to find it.

My first attempt was to just dig around and here's what I got:

```javascript
Object.fromEntries(selectedPlugin.collection.items[0].links.map((o: any) => [o.rel, o.href])).files
```

Next, I consulted the source code of [@fnndsc/chrisapi](https://www.npmjs.com/package/@fnndsc/chrisapi) and tried using its helper methods:

```javascript
import { Collection } from "@fnndsc/chrisapi";
// -- snip --
Collection.getLinkRelationUrls(selectedPlugin.collection.items[0], 'files')[0]
```

The code is pretty complicated but in my opinion, it's preferable to hard-coding. Rather, the fault lies in the implementation of \@fnndsc/chrisapi for not having a getter helper function for this piece of data.

jennydaman (changes requested):

jennydaman:
Personally I am pretty excited about this feature! I hope that you'll continue to work on it.

src/components/feed/NodeDetails/NodeDetails.tsx
Quintinus98: Good insight @jennydaman 
We have made an update, please look at it. Thank you.

src/components/feed/NodeDetails/NodeDetails.tsx
```diff
@@ -221,8 +234,42 @@ const NodeDetails: React.FC<INodeProps> = ({ expandDrawer }) => {
...
+                  aria-label="Download command popover"
+                  position="bottom"
+                  headerContent={<div>chrs Download Files</div>}
+                  hasAutoWidth
+                  bodyContent={
+                    <div>
+                      <p>Copy and Paste the URL below into your chrs cli to download the files</p>
+                      <div className="pf-c-clipboard-copy">
+                      <div className="pf-c-clipboard-copy__group">
+                      <input className="pf-c-form-control" readOnly type="text" value={chrsCommandPath} aria-label="Readonly input example" />
```
jennydaman, line 248: Prefer to use the React component from `@patternfly/react-core` instead of HTML elements + `className`.

https://www.patternfly.org/v4/components/clipboard-copy/#read-only

src/components/feed/NodeDetails/NodeDetails.tsx
```diff
@@ -94,6 +99,14 @@ const NodeDetails: React.FC<INodeProps> = ({ expandDrawer }) => {
     instanceParameters,
     pluginParameters,
   ]);
+  
+  const path = selectedPlugin?.collection.items[0].links.find((obj:any) => (obj.rel === "files")).href
```
jennydaman, line 103: the variable name here is a little misleading. It's not a path, it's an API URI right?

src/components/feed/NodeDetails/NodeDetails.tsx
```diff
@@ -221,8 +234,42 @@ const NodeDetails: React.FC<INodeProps> = ({ expandDrawer }) => {
...
             </Button>
+            <Popover_patternfly
+                  id="node-details-popover"
+                  aria-label="Download command popover"
+                  position="bottom"
+                  headerContent={<div>chrs Download Files</div>}
+                  hasAutoWidth
+                  bodyContent={
+                    <div>
+                      <p>Copy and Paste the URL below into your chrs cli to download the files</p>
```
jennydaman, line 245: The wording doesn't make much sense here, specifically "into your chrs cli" could be reworded. How about more simply:

> Run the command below in a terminal to download the files.

src/components/feed/NodeDetails/NodeDetails.tsx
jennydaman: Be more careful about leaving behind example code: `aria-label="Readonly input example"`

jennydaman (changes requested):

src/components/feed/NodeDetails/NodeDetails.tsx
ciradu2204: Thanks, we have updated it.

Quintinus98:
Hello @jennydaman, we have made the requested changes. Kindly review the PR.
Thank you.

jennydaman:
Closing because #782 was accepted.

PR #682, opened 2022-10-13
add additional loading condition for plugin parameters

### LINKED ISSUE #619: Plugin instance parameter selection loading race condition
priority: low

screenshot

The response from `api/v1/plugins/N/parameters/` was slow for a stochastic reason. The wizard UI was interactable before the parameters data was loaded, so it said "You cannot more parameters to this plugin" but then a few seconds later, the (required) parameters popped up.

https://github.com/FNNDSC/ChRIS_ui/blob/59adf160b570e151b38e59e6a866e7b22ff2a9a4/src/components/feed/AddNode/GuidedConfig.tsx#L96

More technically, there are other possible reasons for `params` to be falsy here other than it representing an empty collection.

BTW let's also fix the English of this error message...

codeEmpress1:
### What this PR does.
- fixes #619 
I found out the error occurs mostly when the internet connection is slow and the plugin params have not been fetched before the UI renders. I fixed this by updating the conditions.
image

### DIFF
--- a/src/components/feed/AddNode/GuidedConfig.tsx
+++ b/src/components/feed/AddNode/GuidedConfig.tsx
@@ -87,7 +87,7 @@ const GuidedConfig = ({
       });
     }
 
-    if (params && count >= params.length) {
+    if (params && params.length !== 0 && count > params.length - 1) {
       setConfigState({
         ...configState,
         errors: ["You cannot add more parameters to this plugin"],


src/components/feed/AddNode/GuidedConfig.tsx
```diff
@@ -87,7 +87,7 @@ const GuidedConfig = ({
       });
     }
 
-    if (params && count >= params.length) {
+    if (params && params.length !== 0 && count > params.length - 1) {
```
jennydaman, line 90: The problem is that, the HTTP request which populates the variable `params` happens asynchronously. During the time the HTTP request is happening, the UI renders itself, and _(on a slow internet connection)_ it could render before data is available, so that the UI will display "You cannot add more parameters to this plugin", which is not true, in reality, the UI should be displaying "Plugin parameters are loading..." or some other kind of loading bar.

Could you explain to me how your change fixes the race condition?

codeEmpress1: @jennydaman What I did here is` params.length !== 0` was checking if the length of the params array is not zero which means params have been fetched and `count > params.length - 1` checks if the total number of params added (count) is greater than the total number of params. This way the error shows only when params have been fetched and all params have been added.

PR #694, opened 2022-10-13
Create account button modification

Seundede:
The create account button is disabled when the required fields are empty. The disabled attribute can be set to keep a user from clicking on the button until some other condition has been met (email and password field).

Before:
Screenshot 2022-10-13 at 21 33 55

After:
Screenshot 2022-10-13 at 21 33 12


jennydaman (changes requested):
Thanks for submitting this change.

PRs are easier to merge when only code that is relevant to the feature being implemented is changed. Would you please recreate the commit with only `src/pages/SignUp/components/SignupForm.tsx` staged, leaving `src/pages/DataLibrary/components/PACSLookup/QueryBuilder.tsx` and `src/components/feed/CreateFeed/createfeed.scss` out? If there are issues with those latter two files, please fix them in another PR.

PR #699, opened 2022-10-14
Fix height of tabs in sidebar scaling with screen height

### LINKED ISSUE #662: Height of tabs in sidebar are scaled with screen height
Looks good on short screen

image

Looks weird on tall screen

image

Sidebar styling should be adjusted so height of sidebar tabs is reasonable for tall screens.

This is not a Patternfly bug. Consider Cockpit, another Patternfly web app developed by Red Hat:

image

Chizycodes:
Fixes #662 

Before: 
Screenshot from 2022-10-14 03-13-57

After:
Screenshot from 2022-10-14 03-17-26

### DIFF
--- a/src/assets/scss/layout/_layout.scss
+++ b/src/assets/scss/layout/_layout.scss
@@ -33,3 +33,18 @@ section {
     margin-right:1rem !important;
     top:-0.5 !important;
 }
+
+element.style {}
+
+.pf-c-nav__link,
+.pf-c-nav__link:hover,
+.pf-c-nav__link:focus,
+.pf-c-nav__link:active {
+    width: 100%;
+    text-decoration: none;
+    border: none;
+}
+
+.pf-c-nav__link {
+    min-height: 2rem;
+}
--- a/src/assets/scss/layout/_layout.scss
+++ b/src/assets/scss/layout/_layout.scss
@@ -19,9 +19,6 @@ article {
     margin: 1em auto;
     padding: 2em 4em;
     width: 100%;
-    @media (max-width: 560px) {
-        padding: 2em;
-    }
 }
 
 section {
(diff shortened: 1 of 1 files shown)


jennydaman (changes requested):

src/assets/scss/layout/_layout.scss
```diff
@@ -36,3 +36,18 @@ section {
...
+.pf-c-nav__link:hover,
+.pf-c-nav__link:focus,
+.pf-c-nav__link:active {
+    width: 100%;
+    text-decoration: none;
+    border: none;
+}
+
+.pf-c-nav__link {
+    min-height: 2rem;
```
jennydaman, line 52: I ran your code and found out that this is the only rule which matters. The above rules applying to `.pf-c-nav__link, .pf-c-nav__link:hover, .pf-c-nav__link:focus, .pf-c-nav__link:active`, _seem_ to have no effect, looks the same if they're commented out. Correct me if I'm wrong... Do they serve a purpose?

src/assets/scss/layout/_layout.scss
```diff
@@ -36,3 +36,18 @@ section {
...
     top:-0.5 !important;
 }
+
+element.style {}
+
+.pf-c-nav__link,
+.pf-c-nav__link:hover,
+.pf-c-nav__link:focus,
+.pf-c-nav__link:active {
+    width: 100%;
```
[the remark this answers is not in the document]
Chizycodes: The sidebar uses vh (viewport height) by default for the heights of the nav links. That's why it increases with the height of your screen.

src/assets/scss/layout/_layout.scss
Chizycodes: I just checked. No, they don't, my bad. I'm sending an update now.

PR #703, opened 2022-10-14
Fix create account button functionality

Seundede:
The create account button is disabled when the required fields are empty. The disabled attribute can be set to keep a user from clicking on the button until some other condition has been met (email and password field).

Before:
Screenshot 2022-10-14 at 13 28 33
After:
Screenshot 2022-10-14 at 13 31 56


jennydaman (approved):

jennydaman:
Thank you for the small and good feature!

PR #708, opened 2022-10-16
login form error messages edited

AndroGenius-codes:
fixes issue 
#690 and #706 

login error solved

Issue #690, opened 2022-10-13
error message is not visible upon login with invalid credentials


nanfuka:
![Screenshot from 2022-10-17 07-25-23](https://user-images.githubusercontent.com/29108523/196089512-9858cecb-6c78-4bcb-9cdf-ab668af4033d.png)
go to the Chris ui login page, enter invalid username and password, click save. the error message will pop up at the top. The error message is not not visible since its the same color as the rest of the text of the page. it should have the resemblance of the rest of the error messages in the application. Also the login button should respond upon being clicked.

Timonwa:
Hi, @nanfuka can you add a screenshot of the error? and if you are already working on this, can you please indicate?

thessakranendonk:
Hi @nanfuka! I'm an Outreachy participant and I looked at the bug. I think I figured out what's causing this to happen. By trying to log in and looking at the issue I also noticed the login button doesn't respond to the enter press. I would like to add that functionality too if that's okay.

nanfuka:
If you look closely at the image above, you will notice an error message at the top, but its not easy to tell whether this is an error message since it differs from other error messages. the issue is it should look like  the other error messages in the application. also when you click on the login button, you might not know that you clicked on it since its appearance remains constant

nanfuka:
may be someone else can take it on. i have not assigned it to myself.

AndroGenius-codes:
Hello there just seeing this. I did a bit of fix for this issue at issue 
#708

thessakranendonk:
Hi sorry I just noticed we both worked on this. I looked at your code and it's interesting to see how we both approached this issue so differently haha. I added the functionality to be able to login by enter press and stored the error messages in an enum.

AndroGenius-codes:
> Hi sorry I just noticed we both worked on this. I looked at your code and it's interesting to see how we both approached this issue so differently haha. I added the functionality to be able to login by enter press and stored the error messages in an enum.

lol oh oh. the more the merrier your fix looks great

jennydaman:
I'm seeing several PRs related to this:

- https://github.com/FNNDSC/ChRIS_ui/pull/702
- https://github.com/FNNDSC/ChRIS_ui/pull/708
- https://github.com/FNNDSC/ChRIS_ui/pull/710

I have skimmed them all and left some comments. I am challenging you all to think of better solutions. In the end, only one (or none) of these PR can get merged, though I will consider all of these attempts in your final Outreachy applications.

PR #710, opened 2022-10-17
Update error message loginform

thessakranendonk:
error message is not visible upon login with invalid credentials #690 & #708 

- Resolved the issues of the Login Error Messages not showing in red
- Login Button is disabled until username and password fields are populated 
- Added functionality for logging in by enter press for the login form button


AndroGenius-codes (commented):
oh this looks brilliant

src/pages/LogIn/components/LoginForm.tsx
```diff
@@ -28,66 +28,76 @@ const LoginFormComponent: React.FC<AllProps> = ({ setAuthToken }: AllProps) => {
...
+      setShowHelperText(true);
 
-      try {
-        token = await ChrisApiClient.getAuthToken(
-          authURL,
-          usernameValue,
-          passwordValue
-        );
-      } catch (error) {
+      if (error && !passwordValue && !usernameValue) {
```
jennydaman, line 54: I am curious, is there any declarative way to do this kind of validation instead of using procedural programming? I have heard of in native HTML there is [form validation](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation). I challenge you to give this some thought.

thessakranendonk: Thank you!! Will work on this!

thessakranendonk: I have been working on the Login form some more but I'm curious what you mean by using native HTML to this particular form. I know how HTML works with putting validation directly on the input fields but this form is made with Patternfly/react-core. 

I have added that the submit button is disabled until credentials are entered so as a result that means both fields are required. However I used the hook isLoginButtonDisabled as found in Patternfly together with useMemo for the disabling of the button as seen in my last commit. Please let me know if you want me to change anything to native HTML.

jennydaman:
thank you for adding the feature of pressing enter to submit!

Issue #698, opened 2022-10-14
Enforce strict code style to prevent merge conflicts


jennydaman:
Currently, ChRIS_ui uses a permissive canned eslint configuration.

https://github.com/FNNDSC/ChRIS_ui/blob/49193ea8b5a0798c0004ff56ee56a0604d72c706/package.json#L67

The IDE auto-formatters of many contributors will do things like add semicolons, change quotes from single to double, ... making PR diffs hard to read and causing merge conflicts between branches.

This is a feature request for us to use a strict linter and a Github Actions workflow to check code style. Merging of PRs should only be allowed if and only if the code style check passes.

We can steal the eslint configuration from ChRIS_store_ui. See https://github.com/FNNDSC/ChRIS_store_ui/issues/168 and its linked PR https://github.com/FNNDSC/ChRIS_store_ui/pull/229

nanfuka:
@Timonwa can I collaborate with you on this on this issues. i have some experience there.

Timonwa:
No problem @nanfuka. I made a comment on the pull request. Let me tag you on it.

PR #700, opened 2022-10-14
Add strict code style

Timonwa:
@jennydaman, I have added the rules as you did in PR https://github.com/FNNDSC/ChRIS_store_ui/pull/229.
Fixes #698 
- Please confirm if I missed any rule or if I need to remove any.
- Also, there was already a `.eslintrc.js` file, so I didn't create a `.eslintrc` file.

- I ran the `lint script`, which gave many errors and warnings.
lint errors

- Then I ran `lint:fix` to fix the ones that can be fixed automatically.
lint errors2

The app is still working fine 😌

- I also ran `npx prettier --write .`  so the printer rule too can be enforced.

But as I write the pull request, I think I should have asked if I was just to add the rules/config and confirm if the rules are okay before implementing them. Please what should I do about this? Should I create a separate pull request for adding the config and another pull request for implementing the config rules, or should i leave it as i have currently done?

**Side note**: For the errors that were not fixed automatically, can I attempt to fix the errors that I can fix manually?

PR #719, opened 2022-10-21
New Add strict style code minus the formatting

Timonwa:
The previous pull request on Adding Strict style code #700 already formatted all the files which would have caused severe merge conflicts for contributors. I did git reset to go back to the instance before the format was made, but because it was being updated with the master branch, it also undid already merged codes, which made it difficult to know which code changes to keep and delete.

so I created a new branch from the master branch and added the strict code rules again. but did not run the prettier format so there would not be changes to other files and code.

Fixes https://github.com/FNNDSC/ChRIS_ui/issues/698


jennydaman (changes requested):
You have added `yarn.lock`, please remove it.

By the way, about deleting files from a git repo: deleting a file doesn't erase it from history, which can be a good or bad thing. For instance, The content of `.eslintcache` can be found here: https://github.com/FNNDSC/ChRIS_ui/commit/940c0ea4d52b140fd3e37f0d7623197a46642c7d

This could be potentially sensitive/private information. It's preferable for you to remove these from git history (though it won't matter when I merge your PR because I can do a squash commit).

Timonwa:
Thank you @jennydaman. I have removed the yarn.lock file.

JulianaSau:
Hi @Timonwa. I removed the pre-push hook for now since it isn't really necessary. 
Once the codebase has been cleared of all linting errors and code formatting errors, then the hook can be added with the commands `format:check` and `lint:check` to confirm that the code styles are enforced

Timonwa:
@jennydaman @PintoGideon, can you check it and lets us know the next step?

package.json
```diff
@@ -64,10 +66,18 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "lint": "eslint '*/**/*.{js,ts,tsx}' --quiet --fix",
+    "type:check": "tsc",
+    "lint": "eslint */**/*.{js,ts,tsx}",
```
jennydaman, line 70: There should be quotes around the glob expression.

package.json
```diff
@@ -64,10 +66,18 @@
...
+    "lint:fix": "eslint --fix */**/*.{jsx,js,tsx}",
+    "prettier": "prettier . --ignore-path .gitignore",
+    "format": "prettier --write",
```
jennydaman, line 73: I think `npm run format` is superfluous when `npm run format:fix` is also defined

lint-staged.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  '*.{js,ts,tsx}': ['eslint --fix', () => 'tsc-files --noEmit'],
+  '*.{js,ts,tsx,json,css,jsx}': ['npm run format'],
+  '*.js': 'eslint --cache --fix',
```
jennydaman, line 4: Prefer to use `npm run ...` commands instead of calling `eslint` directly.

jennydaman (changes requested):

jennydaman:
I tried your changes locally and it works well. I think we should simplify the npm scripts, because really we only need only one command which "does everything" (`prettier --write ... && estline --fix ..`)

package.json
Timonwa: should I do the same thingfor the `npm run lint` is superfluous when `npm run lint:fix` too?

JulianaSau: I'll work on condensing those commands. Because the `format` and `lint` targets the files in staging and `fix` and `check` scripts were to target global linting and formatting

Timonwa: Hi @JulianaSau, I am just seeing your comment. can you help me check if I have made all the changes as they should be, then correct or fix any if necessary and resolve them?

Issue #654, opened 2022-10-11
 A loader while the "Plugins, Pipelines and Compute" fetches


mightyFZeus:
working on  this

mightyFZeus:
#692 closes this issue

@PintoGideon

PR #723, opened 2022-10-21
loader for plugin, pipeline and compute catalog

mightyFZeus:


mightyFZeus:
closes #654.  A loader shows while plugins, pipeline and compute catalog fetches

PR #725, opened 2022-10-21
making plugin selcetable without deselecting the previous selected one when creating new analysis

### LINKED ISSUE #689: Checkbox selection when creating new analysis
when creating new analysis and you are selecting a plugin, once you select one, you wont be able to select another one until the previous selected plugin is deselected.

mightyFZeus:
Screenshot (105)

fixes #689 Prior to  this, when creating a new analysis, when a plugin is selected, you can't select another one until you deselect the selected one. the image below shows no plugin(select option) is disalbed for selection

@jennydaman

### DIFF
--- a/src/components/catalog/PluginCatalog.tsx
+++ b/src/components/catalog/PluginCatalog.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect } from "react";
 import ChrisAPIClient from "../../api/chrisapiclient";
 import DisplayPage from "./DisplayPage";
+import { Title, EmptyState, EmptyStateIcon, Spinner } from '@patternfly/react-core';
 
 const PluginCatalog = () => {
   const [plugins, setPlugins] = React.useState<any[]>();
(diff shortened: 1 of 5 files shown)


PintoGideon (changes requested):
Could you please remove package-lock.json from the PR ?

mightyFZeus:
Okay

src/components/catalog/PluginCatalog.tsx
```diff
@@ -65,6 +70,17 @@ const PluginCatalog = () => {
     });
   };
 
+  if (loading) {
+    return <>
+      <EmptyState>
+        <EmptyStateIcon variant="container" component={Spinner} />
+        <Title headingLevel='h4'>
```
Seundede, line 77: @PintoGideon @mightyFZeus I think an alternative is to make a reusable component called `Loading` that will be used when the loading state is true instead of repeating this code

jennydaman: Moreover, this code is not related to this PR so it should not be included in the patch.

Issue #697, opened 2022-10-14
Sef-updating build string does not work on Windows


jennydaman:
It's useful to see in the "Overview" tab when ChRIS_ui was last updated, so we use a bash script to print git information and the date at build-time. The script output gets rendered into the JSX. https://github.com/FNNDSC/ChRIS_ui/pull/508

This script doesn't work on Windows. https://github.com/FNNDSC/ChRIS_ui/issues/664#issuecomment-1276339800 Even though we don't intend to support Windows as a development environment, it is more elegant if we can achieve a pure node.js project for cross-platform compatibility and perhaps to better benefit from static code analysis tools.

For the most part, I think `print_version.sh` can be translated from bash to pure node.js, and using cross-platform commands such as `git` and `npm` only.

Quintinus98:
@jennydaman I would like to work on this issue. As discussed in the meeting, I'd go ahead and start, since I'm the first to comment on the issue

ciradu2204:
@jennydaman, can I collaborate with @Quintinus98?

Quintinus98:
I'd like to collaborate with you @ciradu2204

PR #728, opened 2022-10-25
 Self-updating build string does not work on Windows #697 

Quintinus98:
**Description**
It's useful to see in the "Overview" tab when ChRIS_ui was last updated, so we use a bash script to print git information and the date at build-time. This script doesn't work on Windows. It is more elegant if we can achieve a pure node.js project for cross-platform compatibility and perhaps to better benefit from static code analysis tools.

**The Fix**
Translated the print_version.sh file from bash to pure node.js, and using cross-platform commands such as git and npm only.
Screenshot from 2022-10-25 06-46-46

**Outcome**
The _Build_ syntax on the Overview page as seen, now works on pure nodejs and is cross-platform compatible.

**Notes**
Removed:
- print_version.sh file
- print_version.sh script in package.json

**Original Issue**
This PR fixes issue #697


ciradu2204:
@jennydaman, we are waiting for your review. Thank You

src/pages/Dashboard/Dashboard.tsx
```diff
@@ -56,10 +55,40 @@ const DashboardPage = (props: DashboardProps) => {
...
+      const today = new Date();
+      const yyyy = today.getFullYear().toString();
+      const mm = String(today.getMonth() + 1).padStart(2, '0');
+      const dd = String(today.getDate()).padStart(2, '0');
+      return [yyyy, mm, dd].join("");
+    }
+
+    const diff = () => {
+      // Run external tool synchronously
+      const result = shell.exec('git diff --quiet src/').code
```
jennydaman, line 72: Prefer not to use dependency if the standard library function is easy enough to use.

```javascript
child_process.spawnSync('git', ['diff', '--quiet', 'src/']).status
```

jennydaman (changes requested):

src/pages/Dashboard/Dashboard.tsx
Quintinus98: Thanks for the update, I will make the requested changes

Quintinus98:
@jennydaman I have made the requested changes. Could you have a look at it and merge the PR if it meets your requirements.
Thank you.

Issue #713, opened 2022-10-18
Number of feeds in header displayed as [object Object]


jennydaman:
<img width="1375" alt="Screen Shot 2022-10-18 at 4 51 59 PM" src="https://user-images.githubusercontent.com/20404439/196542569-60959d38-21a1-440b-a5d0-f6a290661d23.png">

ademolaomosanya:
i would love to work on this @jennydaman

ademolaomosanya:
okay @dipanshuraz

PR #731, opened 2022-10-26
fix:Number of feeds in header displayed as [object Object] #713

Sally-Builds:
Fixed bug issue #713


Headon1729:
Hey @Sally-Builds 
Not sure but aren't you also removing the styling intended for the component ?

PR #738, opened 2022-11-01
Fix: Modal sizes in the project

Seundede:
This PR fixes the size of some modals in the project

Before:
Untitled design (5)
After:
Untitled design (5)

Before:
Untitled design (7)

### DIFF
--- a/src/components/feed/CreateFeed/createfeed.scss
+++ b/src/components/feed/CreateFeed/createfeed.scss
@@ -21,7 +21,7 @@
 }
 
 .pf-c-modal-box {
-  height: 50%;
+  height: 30%;
 }
 
 .feed-create-wizard {
--- a/src/components/feed/CreateFeed/createfeed.scss
+++ b/src/components/feed/CreateFeed/createfeed.scss
@@ -21,7 +21,7 @@
 }
 
 .pf-c-modal-box {
-  height: 95%;
+  height: 50%;
 }
 
 .feed-create-wizard {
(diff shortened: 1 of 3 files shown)


PintoGideon (changes requested):
Hello there,

We want the modal size in the create wizard to be full screen. Also, the add node modal needs to be a decent height and width to display the content inside the modal well. Maybe, a deeper thought would require in how to adjust the width and heights of the modal throughout the UI. The height and width of the delete node modal looks fine to me. One size doesn't fit all in this case unfortunately.


Screenshot from 2022-11-01 10-51-00

Screenshot from 2022-11-01 10-50-52

Seundede:
Thanks for the feedback. I'll work on it now

Seundede:
@PintoGideon the current changes should fix the issue. I added a style to target the modal on the Analysis page

Seundede:
Will inline-style be better?

PintoGideon (changes requested):
Looks like the changes were reverted back. Now all the modals have the same height as it was before this PR. Is this PR fixing any issue currently ? Could you try inline styles and see if you are able to have different heights for different modals. Let me know if you have a solution to this.

Seundede:
This PR fixes the modal of the `Download selected feeds`,  `Merge selected feeds`, `Duplicate selected feeds` and `delete selected feeds` on the  `new and existing analyses` page.
I undid the first changes I made since they affected the `create new analysis` and `add new node` modals.
Before:
Screenshot 2022-11-01 at 20 18 43
After:
Screenshot 2022-11-01 at 20 18 28

PintoGideon:
Oh I see @Seundede . Thanks for clarifying that. Looks good to me.

PR #744, opened 2022-11-03
Enable scale nodes by output size

oryxwhite:
chrisoutputsize

I've implemented and enabled the output size option.  The scale is based on the size of the output directory of each plugin.  The 
file list is fetched when the component mounts and a reduce function is run, generating the total size of the directory.  In the screenshot you can see the result of a file conversion, a reduction in the directory size by half.

### DIFF
--- a/src/components/feed/FeedTree/Node.tsx
+++ b/src/components/feed/FeedTree/Node.tsx
@@ -1,10 +1,11 @@
-import React, { Fragment, useRef } from 'react'
+import React, { Fragment, useRef, useState } from 'react'
 import { select } from 'd3-selection'
 import { HierarchyPointNode } from 'd3-hierarchy'
 import { Datum, TreeNodeDatum, Point } from './data'
 import { PluginInstance } from '@fnndsc/chrisapi'
 import { useTypedSelector } from '../../../store/hooks'
 import { FeedTreeScaleType } from './Controls'
+import { getPluginFiles } from '../../../store/resources/saga'
 
 type NodeWrapperProps = {
   tsNodes?: PluginInstance[]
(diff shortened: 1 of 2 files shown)


oryxwhite:
@mairin @jennydaman @PintoGideon would you mind reviewing this? thanks

src/components/feed/FeedTree/Node.tsx
```diff
@@ -171,12 +172,36 @@ const NodeMemoed = React.memo(Node)
...
+    })
+
+    if (pluginInstance) {
+      const pluginFiles = getPluginFiles(pluginInstance).next().value
+      if (pluginFiles instanceof Promise<any[]>) {
+        pluginFiles.then((files) =>{
+          const fileSizes = files.map((file) => {
+            return file.collection.items[0].data[3].value
+          })
+          const dirSize = fileSizes.reduce((previousValue, currentValue) => previousValue + currentValue, 0)
```
jennydaman, line 193: yay functional programming! I like it

src/components/feed/FeedTree/Node.tsx
```diff
@@ -171,12 +172,36 @@ const NodeMemoed = React.memo(Node)
...
+  )
+
+  React.useEffect(() => {
+    const pluginInstance = pluginInstances?.find((instance) => {
+      return instance.collection.items[0].data[0].value === data.id
+    })
+
+    if (pluginInstance) {
+      const pluginFiles = getPluginFiles(pluginInstance).next().value
+      if (pluginFiles instanceof Promise<any[]>) {
```
jennydaman, line 188: `instanceof Promise` seems awkward to me. In what situation might this conditional be false? If it's in the case of a network error or server error, could we handle this somehow by showing or logging an error message?

src/components/feed/FeedTree/Node.tsx
```diff
@@ -171,12 +172,36 @@ const NodeMemoed = React.memo(Node)
...
+        pluginFiles.then((files) =>{
+          const fileSizes = files.map((file) => {
```
jennydaman, line 190: does not account for pagination. I checked the helper you're calling and it hard-codes 200 files. We need to think of how to better handle pagination across this entire project, maybe using async iterable

https://github.com/FNNDSC/ChRIS_ui/blob/68b9037c6236828a50df7ac0ca45799dca90896a/src/store/resources/saga.ts#L23-L29

src/components/feed/FeedTree/Node.tsx
```diff
@@ -191,7 +216,9 @@ const NodeWrapper = (props: NodeWrapperProps) => {
       scale = Math.log10(end.getTime() - start.getTime()) / 2
     }
   } else if (overlayScale === 'size') {
-    // props.data.item?.
+      if (outputSize) {
+        scale = outputSize / (1024 ** 2) / 10
```
jennydaman, line 220: the size of nodes will probably need to be scaled relative to the maximum size in the feed.

jennydaman (changes requested):

src/components/feed/FeedTree/Node.tsx
oryxwhite: I originally used a conditional because `getPluginFiles` returns an array of type `any[] | Promise<any[]>` so the typechecker requires the conditional to confirm the array of Promises.  I refactored this to call `getResource` directly, which returns a Promise, handles errors and allows me to limit the network requests to 20 files.

src/components/feed/FeedTree/Node.tsx
oryxwhite: I refactored this to scale to the node window, with the scale maxing out at 700MB.  This seems to be able to represent a range of file sizes, although it's a bit hard to differentiate when the sizes are small (under 50MB or so).  The other option would be to set the max relative to the size of the whole feed, which is what I think you're suggesting, however, we would have to account for changes in the size as more nodes are added, which might be confusing.  Which do you think is better?

oryxwhite:
@jennydaman would you mind reviewing my changes?

Issue #766, opened 2022-12-19
Improve the accessibility of the create analysis flow


ciradu2204:
The user experience when creating a new analysis could be improved to make it simple,  faster, and accessible. The following list shows the different features that need a fix: 

1.  Basic Information
- [x] When a form is opened, add autofocus on the first form field: “Feed Name *”
- [x] When a user hits enter key after entering the feed name/feed description/tags, move to the next step 
- [x]  When a user hits the arrowRight or arrowLeft keys after inserting the needed info, move to the next or prev respectively.
2. Feed Type Selection 
- [x] Converting radio button selection of “Feed type” (Step 2) into [card-based selection](https://design.penpot.app/#/view/35dd4c70-bf14-80a7-8001-c23493ca5292?page-id=35dd4c70-bf14-80a7-8001-c23493ca5293&section=interactions&index=0&share-id=9c5ceddf-c68c-8036-8001-c2361af9a7f4) with icons and heading 
- [x] Change the header on the Feed Type Selection => Feed Data Selection
- [x] Consider replacing the term “Feed” with the term “Analysis”
- [x] When a user hits the g, u, and f keys select the generate, upload or fetch card and move to the next item respectively
- [x] When a user hits the arrowRight or arrowLeft keys after selection, move to the next or prev respectively.
3. Select the FS plugin
- [x] Since the user is allowed to choose only one plugin, radio buttons would work better than checkbox 
- [x] Allow tab selection on radio buttons
- [x]  When a user hits the arrowRight or arrowLeft keys after selection, move to the next or prev respectively.
3. Parameter Configuration
- [x] Add autofocus on the required param
- [x]  Disable the next page if there is a required param not filled
- [x]  When a user hits the arrowRight or arrowLeft keys after inserting the needed param, move to the next or prev respectively.
4. File select from swift
- [x] Allow tab selection on the tree checkbox
- [x]  When a user hits the arrowRight or arrowLeft keys after checking the folder, move to the next or prev respectively.
5. Local select 
- [x]  Allow tab selection on the file upload 
- [x]  When a user hits the arrowRight or arrowLeft keys after selecting a file, move to the next or prev respectively.
6. Pipeline
- [x]  When the user opens up a pipeline expander - it should be selected automatically. Don’t deselect unless the user opens up a different pipeline, in which case the new pipeline that got opened will be chosen and the old one unselected.
[...]

PR #774, opened 2022-12-29
Fix/improve user experience

ciradu2204:
This pr solves the accessibility and user experience issues of the create analysis flow mentioned in #766.


PintoGideon:
@ciradu2204  LGTM. Tested all of the features above. The code looks good to me too.

Great work here !!

PR #782, opened 2023-01-10
Add copy path button

jennydaman:
image


PintoGideon:
Hello @jennydaman  This works as expected. I'll update `@patternfly-react-core` to it's latest version too. Thanks for the contribution.

PR #793, opened 2023-01-13
Indicate successful copy in tooltip of copy button in feed output file browser

jennydaman:
When clicking on the copy button, the tooltip's text will change briefly.


PintoGideon:
Thanks @jennydaman

PR #801, opened 2023-01-16
Improve optional params feature

### LINKED ISSUE #775: Improve the Guided Config component: adding parameters to a plugin instance
- The add parameter button should be removed 
- When a user opens the Guided Config page, it should automatically add a default dropdown
- Once a user selects a param from the dropdown and enters a value, a new dropdown should be added 
- The dropdown should be open by default and show only params which are not yet used.

ciradu2204:
fixes #775 

Before: 
<video src="https://user-images.githubusercontent.com/37863089/212745665-562c4000-dd0e-4e68-88dc-7a80a284132d.mov"  width="180" />

After: 
<video src= "https://user-images.githubusercontent.com/37863089/212746600-1e11bbed-fa7b-4238-ad08-e5b3b37b37cd.mov" width= "180"/>


jennydaman (changes requested):
Looking good so far, thank you for the screen recording.

There is a regression to the handling of boolean parameters. When a boolean parameter is specified, the input field for the parameter value should not be editable.

Here's what it looks like on https://app.chrisproject.org:

image

(This screenshot isn't correct either, because when a boolean flag is specified, the opposite of its default value should be shown, _not_ its default value)

ciradu2204:
> Looking good so far, thank you for the screen recording.
> 
> There is a regression to the handling of boolean parameters. When a boolean parameter is specified, the input field for the parameter value should not be editable.
> 
> Here's what it looks like on https://app.chrisproject.org:
> 
> image
> 
> (This screenshot isn't correct either, because when a boolean flag is specified, the opposite of its default value should be shown, _not_ its default value)

okay, I will have a look at it again

PintoGideon:
Hello @ciradu2204 , Great job on this feature. 

I found a bug. When I hit 'Enter' on required parameters, I see that the node get's added to the parent and the modal is closed. 


https://user-images.githubusercontent.com/15992276/213209388-ed571f97-d894-4c5e-bdc5-4f57d27fb91a.mp4

ciradu2204:
> Hello @ciradu2204 , Great job on this feature.
> 
> I found a bug. When I hit 'Enter' on required parameters, I see that the node get's added to the parent and the modal is closed.
> 
>  Required_Parameters_Bug.mp4

oooh okay, let me have a look at it

Issue #784, opened 2023-01-10
Ugly hash URL appears when clicking on filebrowser breadcrumb in feed view


jennydaman:
known affected version 3ebef7788ec0ac9f9cb4cf10129ff8f6e368767a

![image](https://user-images.githubusercontent.com/20404439/211626467-e7fa79b0-da89-4ff2-b5d8-81341f708947.png)

If you click on "subjects" in the breadcrumbs, the URL changes from `http://localhost:3000/feeds/1` to `http://localhost:3000/feeds/1#`

betty-alagwu:
Hello @jennydaman i've been able to reproduce this bug, i'll love to work on it

betty-alagwu:
Thank you, on it

PR #807, opened 2023-01-22
fix the hash URL on the fileBrowser breadcrumb

betty-alagwu:
Fixes:  #784

### DIFF
--- a/src/components/feed/FeedOutputBrowser/FileBrowser.tsx
+++ b/src/components/feed/FeedOutputBrowser/FileBrowser.tsx
@@ -154,7 +154,7 @@ const FileBrowser = (props: FileBrowserProps) => {
         showDivider={true}
         key={index}
         onClick={onClick}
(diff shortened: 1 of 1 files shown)


jennydaman:
@PintoGideon according to the `git blame` you added the `#` there, do you remember what its purpose was?

PintoGideon:
Hey @jennydaman , Unfortunately I don't remember why I added it.

betty-alagwu:
@jennydaman If there is any changes you want me to implement, i will be happy to do so.

jennydaman:
@PintoGideon we should think about this a bit.

@betty-alagwu I'm trying to think of reasons why we would want to add a hash symbol into the URL. Sometimes an undesirable reload happens if you don't add the hash symbol. Would you describe to me how you've tested this change? If possible, it  would be best to add some automatic tests covering the clicking of this breadcrumb.

Since the URL is not supposed to change at all when using this breadcrumb, what is the purpose of the `to` prop anyway?

PR #817, opened 2023-02-01
add functionality to disable button and remove edit

### LINKED ISSUE #714: Disable "Add a Child Node" button and edit title functionality if logged in user does not own the feed
Screen Shot 2022-10-18 at 4 54 05 PM

Logged in as the user "chris" but feed owner is "Rudolph". Cannot edit feeds owned by another user, so any edit operations should be disabled/greyed out.

itsgurmannatsohal:
After:
image
Before:
image

fixes #714

### DIFF
--- a/src/components/feed/AddNode/AddNode.tsx
+++ b/src/components/feed/AddNode/AddNode.tsx
@@ -17,6 +17,8 @@ import { getRequiredObject } from "../CreateFeed/utils/createFeed";
 import "./styles/AddNode.scss";
 import { useTypedSelector } from "../../../store/hooks";
 import { useDispatch } from "react-redux";
+import { Cookies } from 'react-cookie'
+
 
 function getInitialState() {
   return {
@@ -33,12 +35,24 @@ function getInitialState() {
 }
 
 const AddNode: React.FC<AddNodeProps> = ({
+  readOnly,
   selectedPlugin,
   pluginInstances,
   getParams,
   addNode,
   params,
 }: AddNodeProps) => {
+  const currentFeedPayload = useTypedSelector((state) => state.feed.currentFeed)
+  const cookie = new Cookies()
+  const user2 = cookie.get('username')
+  const { data: feed} = currentFeedPayload;
+  if(feed && feed.data.creator_username===user2){
+    readOnly = false
+  } else {
+    readOnly = true
+  }
+
+
   const dispatch = useDispatch();
   const [addNodeState, setNodeState] =
     React.useState<AddNodeState>(getInitialState);
@@ -340,12 +354,21 @@ const AddNode: React.FC<AddNodeProps> = ({
 
   return (
     <React.Fragment>
-      <Button icon={<MdOutlineAddCircle />} type="button" onClick={toggleOpen}>
+      {readOnly ? (
+        <div className="tooltip-container">
+        <button className="tooltip-hoverable" disabled={true} >
+        {<MdOutlineAddCircle />} Add a Child Node{" "}( C )
+        </button>
+        <span className="tooltip">You are not the feed owner</span>
+        </div>
+      ) : (
+        <Button icon={<MdOutlineAddCircle />} type="button" onClick={toggleOpen}>
         Add a Child Node{" "}
         <span style={{ padding: "2px", color: "#F5F5DC", fontSize: "11px" }}>
           ( C )
         </span>
       </Button>
+      )}
       {childNode && (
         <Wizard
           isOpen={childNode}
--- a/src/components/feed/AddNode/AddNode.tsx
+++ b/src/components/feed/AddNode/AddNode.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { Dispatch } from "redux";
 import { connect } from "react-redux";
-import { Wizard, Spinner, Button } from "@patternfly/react-core";
+import { Wizard, Spinner, Button, Tooltip } from "@patternfly/react-core";
 import { MdOutlineAddCircle } from "react-icons/md";
 import GuidedConfig from "./GuidedConfig";
 import Editor from "./Editor";
@@ -356,10 +356,16 @@ const AddNode: React.FC<AddNodeProps> = ({
     <React.Fragment>
       {readOnly ? (
         <div className="tooltip-container">
-        <button className="tooltip-hoverable" disabled={true} >
+          <Tooltip content={
+          <div>
+            You are not the feed owner
+          </div>
+        }
+          >
+        <Button isAriaDisabled variant="secondary">
         {<MdOutlineAddCircle />} Add a Child Node{" "}( C )
-        </button>
-        <span className="tooltip">You are not the feed owner</span>
+        </Button>
+        </Tooltip>
         </div>
       ) : (
         <Button icon={<MdOutlineAddCircle />} type="button" onClick={toggleOpen}>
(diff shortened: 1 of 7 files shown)


src/pages/Feeds/Feed.scss
```diff
@@ -7,166 +7,173 @@
...
-  to {
-    width: 0;
-    height: 0;
-    visibility: hidden;
-  }
-}
-.feed_modal {
-  height: 30%;
-}
+ @import "../../assets/scss/helpers/mixins";
```
jennydaman, line 10: for what reason is this diff so large?

src/components/feed/AddNode/AddNode.tsx
```diff
@@ -33,12 +35,24 @@ function getInitialState() {
 }
 
 const AddNode: React.FC<AddNodeProps> = ({
+  isCreator,
```
jennydaman, line 38: nitpick: it would be better to use a variable name such as `isOwner`, or perhaps even better, `readOnly`.

The creator of a feed might not necessarily have permission to click the buttons, it is the feed owner who has those permissions, and feed ownership can be changed.

I think a variable name `readOnly` is more clear for what the purpose of the variable is, and it is general so it leaves room for future features. For instance, one day we might want to implement some toggle for "Editing/Reading" mode similar to Google Docs and Vim.

src/components/feed/AddNode/AddNode.tsx
```diff
@@ -340,12 +354,18 @@ const AddNode: React.FC<AddNodeProps> = ({
...
     <React.Fragment>
+      {isCreator ? (
       <Button icon={<MdOutlineAddCircle />} type="button" onClick={toggleOpen}>
         Add a Child Node{" "}
         <span style={{ padding: "2px", color: "#F5F5DC", fontSize: "11px" }}>
           ( C )
         </span>
       </Button>
+      ) : (
+      <button disabled={true} >
```
jennydaman, line 365: When a button is disabled, it is good for there to be a popup/tooltip explaining why.

src/components/feed/NodeDetails/PluginTitle.tsx
```diff
@@ -81,11 +92,15 @@ const PluginTitle = () => {
           <Title headingLevel="h3" size="xl">
             <span>{pluginName}</span>
           </Title>
-          <AiFillEdit
+          {isCreator ? (
```
jennydaman, line 95: same advice as above, do you think you could add a tooltip here explaining why the edit feature is not available?

src/components/feed/AddNode/AddNode.tsx
jennydaman, line 365: We have a sequence of buttons to render. Currently, they are:

- Add a Child Node (rw)
- Add a Pipeline (rw)
- Download Files (ro)
- View Terminal (ro)
- Add a Graph Node (rw)
- Delete Node (rw)

rw meaning "read-write" permissions needed, ro meaning "read-only" permissions. The feed owner has rw permissions, users who the feed has been shared with (but who do not own the feed) have ro permissions.

We should try to generalize the conditional that a button should be disabled and apply the solution to the other rw buttons.

jennydaman (changes requested):

src/pages/Feeds/Feed.scss
itsgurmannatsohal: The actual difference in only in the last 6 lines. The rest has a little difference in indentation, mistakenly. Shall I revert that or is it alright?

src/components/feed/AddNode/AddNode.tsx
itsgurmannatsohal: Okay

src/components/feed/NodeDetails/PluginTitle.tsx
itsgurmannatsohal: Sure, I'll add that

src/components/feed/AddNode/AddNode.tsx
itsgurmannatsohal: Shall I use jQuery UI library?

itsgurmannatsohal: I'll try generalizing this, but I might need a little guidance.

itsgurmannatsohal: Would the message "You are not the feed owner" be okay?

src/components/feed/AddNode/AddNode.tsx
```diff
@@ -340,12 +354,21 @@ const AddNode: React.FC<AddNodeProps> = ({
 
   return (
     <React.Fragment>
-      <Button icon={<MdOutlineAddCircle />} type="button" onClick={toggleOpen}>
+      {readOnly ? (
+        <div className="tooltip-container">
+        <button className="tooltip-hoverable" disabled={true} >
+        {<MdOutlineAddCircle />} Add a Child Node{" "}( C )
+        </button>
+        <span className="tooltip">You are not the feed owner</span>
```
jennydaman, line 362: Use patternfly's implementation of tooltip instead of a manual implementation.

https://www.patternfly.org/v4/components/button#aria-disabled-examples

itsgurmannatsohal: Done. Please review

src/components/feed/AddNode/AddNode.tsx
```diff
@@ -33,12 +35,24 @@ function getInitialState() {
...
   getParams,
   addNode,
   params,
 }: AddNodeProps) => {
+  const currentFeedPayload = useTypedSelector((state) => state.feed.currentFeed)
+  const cookie = new Cookies()
+  const user2 = cookie.get('username')
+  const { data: feed} = currentFeedPayload;
+  if(feed && feed.data.creator_username===user2){
+    readOnly = false
```
jennydaman, line 50: if `readOnly` is a prop, then it should not be modified. To elaborate:

- if `readOnly` is a prop, then it is the parent's responsibility to evaluate the logic `feed.data.creator_username===user2`
- if `readOnly` is a state, then it is this component's responsibility to evaluate the logic `feed.data.creator_username===user2`

src/components/feed/AddNode/AddNode.tsx
```diff
@@ -340,12 +354,27 @@ const AddNode: React.FC<AddNodeProps> = ({
 
   return (
     <React.Fragment>
-      <Button icon={<MdOutlineAddCircle />} type="button" onClick={toggleOpen}>
+      {readOnly ? (
+        <div className="tooltip-container">
```
jennydaman, line 358: Please re-indent your code.

src/components/feed/NodeDetails/PluginTitle.tsx
```diff
@@ -1,16 +1,28 @@
...
   return selectedPlugin?.data.title || selectedPlugin?.data.plugin_name
 }
 
-const PluginTitle = () => {
+const PluginTitle: React.FC<AddNodeProps>  = ( {readOnly}) => {
+  const currentFeedPayload = useTypedSelector((state) => state.feed.currentFeed)
+  const cookie = new Cookies()
+  const user2 = cookie.get('username')
+  const { data: feed} = currentFeedPayload;
+  if(feed && feed.data.creator_username===user2){
```
jennydaman, line 21: duplication of code. See previous comment about prop v.s. state.

If you see that each button needs to decide `readOnly` for themselves, then a better solution would be to move the logic to the parent which contains all the buttons.

PR #833, opened 2023-02-09
Replace docker run with apptainer exec

### LINKED ISSUE #730: Replace `docker run` command with `apptainer exec` usage
Please compare/constrast the usage between `docker run` and `apptainer exec` then replace the suggestion of using `docker run` with `apptainer exec`.

https://github.com/FNNDSC/ChRIS_ui/blob/97734946828b486d84144e1c39362b8be642854e/src/components/feed/NodeDetails/NodeDetails.tsx#L322

itsgurmannatsohal:
fixes #730

### DIFF
--- a/src/components/feed/NodeDetails/NodeDetails.tsx
+++ b/src/components/feed/NodeDetails/NodeDetails.tsx
@@ -352,7 +352,7 @@ function getCommand(
     }
   }
 
-  let command = `$> docker run --rm -v $(pwd)/in:/incoming -v $(pwd)/out:/outgoing \\\n${dock_image} ${selfexec} `;
+  let command = `$> docker run -d -v $(pwd)/in:/incoming -v $(pwd)/out:/outgoing ${dock_image} \\\n$ container_id=$(docker ps -q --filter "ancestor=${dock_image}") \\\n$ docker exec $container_id ${selfexec}\\\n $docker stop $container_id \\\n`;
   let parameterCommand = [];
 
   if (modifiedParams.length) {
--- a/src/components/feed/NodeDetails/NodeDetails.tsx
+++ b/src/components/feed/NodeDetails/NodeDetails.tsx
@@ -352,7 +352,7 @@ function getCommand(
     }
   }
 
-  let command = `$> apptainer exec --rm -v $(pwd)/in:/incoming -v $(pwd)/out:/outgoing \\\n${dock_image} ${selfexec} `;
+  let command = `$> apptainer oci delete ${dock_image} & apptainer exec --bind $(pwd)/in:/incoming,$(pwd)/out:/outgoing ${dock_image} ${selfexec}`;
   let parameterCommand = [];
 
   if (modifiedParams.length) {
--- a/src/components/feed/NodeDetails/NodeDetails.tsx
+++ b/src/components/feed/NodeDetails/NodeDetails.tsx
@@ -352,7 +352,7 @@ function getCommand(
     }
   }
 
-  let command = `$> apptainer oci delete ${dock_image} & apptainer exec --bind $(pwd)/in:/incoming,$(pwd)/out:/outgoing ${dock_image} ${selfexec}`;
+  let command = `$> apptainer exec --bind $PWD/in:/incoming,$PWD/out:/outgoing docker://${dock_image} ${selfexec}`;
   let parameterCommand = [];
 
   if (modifiedParams.length) {
--- a/src/components/feed/NodeDetails/NodeDetails.tsx
+++ b/src/components/feed/NodeDetails/NodeDetails.tsx
@@ -352,7 +352,7 @@ function getCommand(
     }
   }
 
-  let command = `$> docker run -d -v $(pwd)/in:/incoming -v $(pwd)/out:/outgoing ${dock_image} \\\n$ container_id=$(docker ps -q --filter "ancestor=${dock_image}") \\\n$ docker exec $container_id ${selfexec}\\\n $docker stop $container_id \\\n`;
+  let command = `$> apptainer exec --rm -v $(pwd)/in:/incoming -v $(pwd)/out:/outgoing \\\n${dock_image} ${selfexec} `;
   let parameterCommand = [];
 
   if (modifiedParams.length) {


src/components/feed/NodeDetails/NodeDetails.tsx
```diff
@@ -352,7 +352,7 @@ function getCommand(
     }
   }
 
-  let command = `$> docker run --rm -v $(pwd)/in:/incoming -v $(pwd)/out:/outgoing \\\n${dock_image} ${selfexec} `;
+  let command = `$> docker run -d -v $(pwd)/in:/incoming -v $(pwd)/out:/outgoing ${dock_image} \\\n$ container_id=$(docker ps -q --filter "ancestor=${dock_image}") \\\n$ docker exec $container_id ${selfexec}\\\n $docker stop $container_id \\\n`;
```
jennydaman, line 355: @itsgurmannatsohal could you explain to me what the `apptainer exec` command is and how it differs from `docker run`?

itsgurmannatsohal: `apptainer exec` is a command used to manage containers and images within an organization. The `apptainer exec` command allows you to run a command within an existing container.

The `docker run` command is used to start a new container from an image. When we use `docker run`, a new container is created, and the command specified is executed within the container.

jennydaman: Hmm, I don't see `apptainer exec` anywhere in your patch.

Apptainer is a completely separate and different container runtime from docker. See https://apptainer.org

The command-line usage of `apptainer exec` and `docker run` are slightly different. In some cases it's preferable to use `apptainer` instead of `docker` because:

- it is usually preinstalled on scientific compute cluster machines
- it does not require you to be the root user by default
- it has weaker filesystem isolation by binding (i.e. mounting as volumes) some host paths automatically, so that the container runs as your user and home directory (among other benefits)


Please compare and contrast `apptainer` from `docker` and then translate this `docker run` command to an `apptainer exec` command. A stretch goal to do here would be adding additional flags as necessary which specify the resource limits, such as using the `--nv` flag if the plugin instance's `gpu_limit` is greater than 0.

itsgurmannatsohal: Sorry I misunderstood the code a bit. I've made the changes, please review.

src/components/feed/NodeDetails/NodeDetails.tsx
```diff
@@ -352,7 +352,7 @@ function getCommand(
     }
   }
 
-  let command = `$> docker run --rm -v $(pwd)/in:/incoming -v $(pwd)/out:/outgoing \\\n${dock_image} ${selfexec} `;
+  let command = `$> apptainer exec --rm -v $(pwd)/in:/incoming -v $(pwd)/out:/outgoing \\\n${dock_image} ${selfexec} `;
```
jennydaman, line 355: There are more differences between the usages of `docker run` and `apptainer exec` which have not been changed here. For instance, if I were to try running the command you provided, here is what I get:

```
Error for command "exec": unknown flag: --rm
```

Please re-review the documentation for `docker run` and `apptainer exec` and then fix the command accordingly. Let me know if you need help with that.

itsgurmannatsohal: I had a doubt. How can I pause the delete command while the processes are being executed. Apptainer doesn't have a command similar to --rm (as far as I could see), so how can I run `${selfexec}` and other commands before delete or is Apptainer procedural and handle that on its own?

itsgurmannatsohal: And I have 2 commands, one that would replace --rm and one for -v. Do I need to make them into one (I tried but I keep getting errors) or two would work?

jennydaman: > is Apptainer procedural [...]

I do not understand what you mean by that.

`--bind` is the appropriate equivalent of `-v`,

itsgurmannatsohal: Yes, I have made the changes in the code accordingly.

jennydaman: Background: a container is not the same thing as an image (a.k.a. container image, docker image). A container is an instance of an image.

A difference between Docker and Apptainer is that, in Docker, the following container states are valid:

- created but not started
- created and finished

In Apptainer, these two states do not exist. Apptainer containers must be created AND running. If containers are not running, then they do not exist.

With the `docker run` command, the `--rm` flag is necessary because otherwise upon finishing the container will enter the "created and finished" state. The desired final state is for the container to have finished and no longer exist.

Since Apptainer containers disappear as soon as they're finished, there is no equivalent to the `--rm` option (it happens automatically).

src/components/feed/NodeDetails/NodeDetails.tsx
```diff
@@ -352,7 +352,7 @@ function getCommand(
     }
   }
 
-  let command = `$> docker run --rm -v $(pwd)/in:/incoming -v $(pwd)/out:/outgoing \\\n${dock_image} ${selfexec} `;
+  let command = `$> apptainer oci delete ${dock_image} & apptainer exec --bind $(pwd)/in:/incoming,$(pwd)/out:/outgoing ${dock_image} ${selfexec}`;
```
jennydaman, line 355: - `apptainer oci delete` is not a necessary step.
- `&` is invalid syntax here. If you mean to run commands one after another, the syntax is `&&`

jennydaman, line 355: The usage of `--bind` looks correct. Though I have a preference for `$PWD` instead of `$(pwd)`, would you change that?

jennydaman, line 355: https://apptainer.org/docs/user/main/cli/apptainer_exec.html#synopsis

The first positional argument of `apptainer exec` should have the prefix `docker://`

jennydaman (changes requested):

src/components/feed/NodeDetails/NodeDetails.tsx
itsgurmannatsohal: Yes, sure

itsgurmannatsohal: Why is `apptainer oci delete` not necessary? We need to delete the container after execution as `--rm` does.

jennydaman (approved):

PR #898, opened 2023-03-29
enables searching plugins by name,title and category

DevBramz:


PintoGideon:
Hey @DevBramz  Could you please provide descriptions in your PR ? I would like to see video snapshots of before and after, api documentation you referred to etc...

PR #906, opened 2023-03-31
fixed unreadable options menu

### LINKED ISSUE #885: Items per page dropdown menu in Pipelines view is unreadable with dark theme
image

DevBramz:
@PintoGideon @jennydaman  ! This PR fixes the styling of item per page drop-down menu in plugins which was unreadable   #885 
Before 
Screenshot from 2023-03-31 20-20-50

After

228562248-fdccc003-8bd1-44dc-9ab0-c6731de6cb4f

### DIFF
--- a/src/components/feed/AddNode/styles/AddNode.scss
+++ b/src/components/feed/AddNode/styles/AddNode.scss
@@ -8,7 +8,14 @@
     color: inherit;
   }
 }
-
+.pf-c-options-menu__menu {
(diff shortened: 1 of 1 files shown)


jennydaman:
@DevBramz whenever it comes to using tricks like `!important` it's good to include an explanation why it's necessary, either in the PR or preferably in the code comments.

Have you tried looking into why the background color of the dropdown is white? Could this be fixed without using `!important` elsewhere, e.g. fixing a "cascaded" style, or maybe upgrading Patternfly?

PR #908, opened 2023-04-03
Concurrent File Uploads

PintoGideon:
This PR focuses on the concurrent file uploads in the ChRIS_ui. The file upload uses the `Promise.allSettled` api which accepts an iterable (e.g., an array) of promises and returns a new promise that is fulfilled with an array of promise state snapshots when all of the input promises have settled (i.e., either fulfilled or rejected). A promise state snapshot is an object that contains a status property indicating whether the promise was fulfilled or rejected, and a value or reason property containing the fulfillment value or rejection reason, respectively.

[Screencast from 04-03-2023 05:14:17 PM.webm](https://user-images.githubusercontent.com/15992276/229629113-be7105fe-3c22-43c2-afae-3dbdd7d74b05.webm)

### DIFF
--- a/src/api/common/index.ts
+++ b/src/api/common/index.ts
@@ -280,3 +280,66 @@ export function catchError(errorRequest: any) {
     return { error_message: errorRequest };
   }
 }
+
+// A function to limit concurrency using Promise.allSettled.
+export const limitConcurrency = async <T>(
+  limit: number,
+  promises: (() => Promise<T>)[],
+  onProgress?: (progress: number) => void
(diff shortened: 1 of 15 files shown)


jennydaman:
Is there a mechanism for limiting concurrent uploads? Performance typically caps between 3-8, and our deployments of CUBE cannot handle more than 16. (Uploads are error prone the more you do)

I think 4 is a good cap.

PintoGideon:
@jennydaman  I'll look into implementing one :)

PR #909, opened 2023-04-04
enables searching plugins by name,title and category

DevBramz:
This Pr enables plugins to be searched by name, title and category other than name alone 
Before
Screenshot from 2023-04-04 18-45-04
Screenshot from 2023-04-04 18-46-03
After


jennydaman (approved):
LGTM

PR #950, opened 2023-10-02
Allow catalog and plugins to be browsed without logging in

jennydaman:
- Removed `<PrivateRoute>` wrappers from the elements of `/catalog` and `/plugin/*` routes
- Clicking on sidebar nav items when not logged in causes error, so the sidebar is now disabled when not logged in.

image


PintoGideon (approved):
Looks good to me.

PR #977, opened 2023-12-20
Fix eslint installation

jennydaman:


jennydaman:
We could consider switching to https://oxc-project.github.io/docs/guide/usage/linter.html

PintoGideon:
Yes, @jennydaman, I'll try oxc lint and send a PR. I have seen it used in a React project, and it works well.

PR #1104, opened 2024-03-12
Simple "copy chrs login" feature

jennydaman:
image

I did not make this feature fancy at all.

I think this should be reworked so that there is more text and a modal. However, the width of the `<Toolbar />` is fixed. The surrounding code should be reworked a bit to use less custom CSS and more Patternfly-friendly coding patterns.

### DIFF
--- a/src/components/Wrapper/Header.tsx
+++ b/src/components/Wrapper/Header.tsx
@@ -51,7 +51,7 @@ export default function Header(props: IHeaderProps) {
           {brand}
         </MastheadBrand>
       </MastheadMain>
-      <MastheadContent className="mastheadContent">
+      <MastheadContent>
         {pageToolbar}
       </MastheadContent>
     </Masthead>
--- a/src/components/Wrapper/Toolbar.tsx
+++ b/src/components/Wrapper/Toolbar.tsx
@@ -73,7 +73,17 @@ const ToolbarComponent: React.FC<AllProps> = (props: AllProps) => {
     setLogoutSuccess();
   };
 
+  const copyLoginCommand = () => {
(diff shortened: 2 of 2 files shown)


src/components/Wrapper/Header.tsx
```diff
@@ -51,7 +51,7 @@ export default function Header(props: IHeaderProps) {
           {brand}
         </MastheadBrand>
       </MastheadMain>
-      <MastheadContent className="mastheadContent">
```
jennydaman, line 54: I deleted this `className` because it was unused.

src/components/Wrapper/Toolbar.tsx
```diff
@@ -73,7 +73,17 @@ const ToolbarComponent: React.FC<AllProps> = (props: AllProps) => {
...
+  const copyLoginCommand = () => {
+    let url = `${window.location.protocol}://${window.location.host}`;
+    let command = `chrs login --ui "${url}" --cube "${import.meta.env.VITE_CHRIS_UI_URL}" --username "${username}" --token "${token}"`;
+    navigator.clipboard.writeText(command);
+    onDropdownToggle();
+  };
+
   const userDropdownItems = [
+    <DropdownItem key="copy-chrs-login" onClick={copyLoginCommand}>
+      CLI login
```
jennydaman, line 85: Text must not be long or else it gets cut off by the right viewport boundary. Dropdown is not able to position itself correctly due to the use of custom CSS.

jennydaman:
Inspiration for a "Copy login command" comes from here:

https://docs.openshift.com/container-platform/4.15/cli_reference/openshift_cli/getting-started-cli.html

Issue #1107, opened 2024-03-13
Possible re-implementation of the PACS Q/R


PintoGideon:
```
git clone --branch=oxidicom https://github.com/FNNDSC/miniChRIS-docker.git
cd miniChRIS-docker
./minichris.sh
docker compose --profile pacs up -d
```

The most important differences are:

1.     The PACS retrieve, repack, register, ... steps are now entirely self-driven. there is no need to poll pfdcm to drive it to completion.
2.     PFDCM is no longer aware of the status of retrieve, repack, register.
3.     DICOM files are registered to CUBE in real-time, meaning they will appear in CUBE within milliseconds of being received from the 
4.     PACS. I.e. to show the progress of a PACS retrieve operation, you should poll CUBE for the number of files in a series, instead of 
        polling pfdcm.

Note that it is still necessary to contact pfdcm to kick things off. pfdcm will:

1.     search for studies and series
2.     tell you how many DICOM instances are in a series
3.     kick off the retrieval operation
 

after telling PFDCM to start doing the retrieval, PFDCM is no longer relevant to the operation.

PintoGideon:
The Series Card component currently lacks responsiveness, particularly on smaller screens. When viewed on devices with limited screen width, the gear icon within the card disappears behind the buttons.

PR #1125, opened 2024-03-22
Implement Pacs Query for oxidicom

PintoGideon:
Addresses: #1107 

This pull request introduces significant enhancements to the PACS retrieval process in ChRIS. The key improvements are as follows:

- Self-driven PACS Operations: The PACS retrieve, push, and register steps are now entirely self-driven. There is no longer a need to poll PFDCM to monitor the progress of these operations. 
- Decoupling from PFDCM: PFDCM is no longer responsible for tracking the status of retrieve, repack, or register operations. 
- Real-time Registration to CUBE: DICOM files are now registered to CUBE in real-time, ensuring they appear within milliseconds of being received from the PACS. To monitor the progress of a PACS retrieve operation, polling CUBE for the number of files in a series is recommended instead of polling PFDCM.
- PFDCM remains relevant only for initiating retrieval operations, searching for studies and series, and providing information about the number of DICOM instances in a series.

These changes streamline the PACS retrieval process, improve real-time feedback, and reduce system complexity.


src/components/Pacs/pfdcmClient.tsx
```diff
@@ -12,27 +12,28 @@ export interface ImageStatusType {
 
 export interface DataFetchQuery {
   SeriesInstanceUID: string;
-  StudyInstanceUID: string;
```
jennydaman, line 15: It turns out that specifying `StudyInstanceUID` is necessary for FUJI PACS. I will fix this.

PR #1162, opened 2024-04-19
add push for staging and and linux/arm64 platform

chhsiao1981:
1. add push for staging.
2. add linux/arm64 platform

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,7 @@ jobs:
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
+
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v3
         with:
@@ -39,7 +40,7 @@ jobs:
       - name: Get postfix of latest
         id: postfix_of_latest
         run: |
-          if [[ $GITHUB_REF_TYPE == tag || $GITHUB_REF_NAME == master || $GITHUB_REF_NAME == staging ]]; then
+          if [[ $GITHUB_REF_TYPE == tag || $GITHUB_REF_NAME == master ]]; then
             echo "Postfix of latest: ''"
             echo "postfix_of_latest=" >> $GITHUB_OUTPUT
           else
@@ -47,7 +48,7 @@ jobs:
             echo "postfix_of_latest=-$GITHUB_REF_NAME" >> $GITHUB_OUTPUT
           fi
 
-      - name: Build
+      - name: Build and export
         uses: docker/build-push-action@v5
         with:
           tags: |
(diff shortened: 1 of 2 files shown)


jennydaman:
Hi @chhsiao1981  I’m happy about adding the staging postfix tag. 

About arm64 builds, the GitHub action runner must do emulation so I worry the build will be very slow (and possibly broken). Since this is a React application, i suggest we do an impure image build:

1. first, build the React project using GitHub Actions outside of docker (always x86_64). This produces HTML/CSS/JS “build asset files”
2. second, copy the “build asset files” into a very minimal container image
3. Install a web server into the container image capable of serving a “single-page application.” Currently, we use sirv-cli

chhsiao1981:
> Hi @chhsiao1981 I’m happy about adding the staging postfix tag.
> 
> About arm64 builds, the GitHub action runner must do emulation so I worry the build will be very slow (and possibly broken). Since this is a React application, i suggest we do an impure image build:
> 
> 1. first, build the React project using GitHub Actions outside of docker (always x86_64). This produces HTML/CSS/JS “build asset files”
> 2. second, copy the “build asset files” into a very minimal container image (any architecture we want)
> 3. Install a web server into the container image capable of serving a “single-page application.” Currently, we use sirv-cli

Hi @jennydaman ～

I believe that the newest update follows your suggestion
that we build the dist in amd64 and copy to both amd64 and arm64.

.github/workflows/build.yml
```diff
@@ -29,24 +29,56 @@ jobs:
...
-            ghcr.io/fnndsc/chris_ui:latest
-            ghcr.io/fnndsc/chris_ui:${{ steps.info.outputs.version_string }}
+            chris_ui:latest${{ steps.postfix_of_latest.outputs.postfix_of_latest}}
           context: .
           file: ./Dockerfile
           platforms: linux/amd64
+          push: false
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          outputs: type=docker,dest=/tmp/chris_ui:latest${{ steps.postfix_of_latest.outputs.postfix_of_latest}}.tar
```
jennydaman, line 62: I have no doubt that this works, but could it be simpler?

There is no need to write the image to a `.tar` then run `docker load ...` There are two approaches which I would prefer.

If we needed to build inside a container, then my preferred solution would be to run a local registry like this:

https://github.com/FNNDSC/microminc/blob/cf6f34685f0a69d933b3a60cb75c7f77b7a17117/.github/workflows/build.yml#L41-L83

However, since ChRIS_ui is a pretty simple React.JS application, I think it would be easier and faster to simply use [`actions/setup-node`](https://github.com/actions/setup-node), build the JavaScript application without using Docker, then build the container image using the `Dockerfile_cp` file you have below.

Dockerfile_cp
```diff
@@ -0,0 +1,21 @@
+# build image
+FROM node:20.7-alpine as builder
+
+WORKDIR /app
+# assuming ./dist is the dist.
+COPY ./dist /app/dist
+
+# production image
+FROM node:20.7-alpine
```
jennydaman, line 9: Multi-stage build is not required because the only instruction we need to execute in the builder image is `COPY ./dist /app/dist`

A multi-stage build (Dockerfile containing more than one `FROM` statement) has the following advantages:

- parallel build
- smaller image size

These two advantages are not applicable to how `Dockerfile_cp` works. 

Would you remove the multistage build?

chhsiao1981: (done)

.github/workflows/build.yml
chhsiao1981: (done)

jennydaman (approved):
I would add some comments to the top of `Dockerfile_cp` to explain that the app should be built on-the-host and then copied into the build container to achieve efficient multiarchitectural build. Everything else looks good to me.

chhsiao1981:
> I would add some comments to the top of `Dockerfile_cp` to explain that the app should be built on-the-host and then copied into the build container to achieve efficient multiarchitectural build. Everything else looks good to me.

(done)

PR #1229, opened 2024-07-22
Local testing

jennydaman:
https://github.com/FNNDSC/miniChRIS-docker is added to this repository as a git submodule, and its startup/shutdown/management commands are added as npm scripts. For example, run `npm run minichris:up` to run miniChRIS, `npm run minichris:down` to shut down miniChRIS, `npm run chrisomatic` to run chrisomatic. Furthermore, scripts for downloading example DICOM data and the `testing/chrisomatic.yml` were added to this repository. This sets up everything we need to run tests of ChRIS_ui locally.

The command `npm run dev:public` was unchanged. The command `npm run dev:local` now does much more. Previously, all it did was run `vite`. Now, the `npm run dev:local` command will (1) run miniChRIS, (2) download sample DICOMs and add them to Orthanc, (3) wipe CUBE's PACSFiles table and delete all DICOM data from CUBE storage.

A new command was added: `npm run test:e2e:local` will run Playwright end-to-end tests, but against a local CUBE instead of using the public testing server. I added a test `tests/pacsQueryAndRetrieve.spec.ts` which only works against the local CUBE. It tests the PACS Q/R page.


PintoGideon (approved):
I was able to get everything up and running and also see the tests in action.

As an aside, I had to run an extra git command to fetch the miniChRIS submodule:

```sh

git checkout -b jennydaman-local-testing master
git pull https://github.com/jennydaman/ChRIS_ui.git local-testing
git submodule update --init --remote --recursive

```

We should point to the miniChRIS documentation in our README so that users can have make and parallel installed as prerequisites.

Inspiring work. Thank you 👍🏽

PR #1240, opened 2024-08-26
Do `npm run test:prepare` in separate GHA step

jennydaman:
This might help with timing and avoiding timeouts when running tests in GitHub Actions.


PintoGideon:
Oh nice.  👍🏽

PR #1251, opened 2024-09-23
simplify build criterion

chhsiao1981:


chhsiao1981:
actually no need to PR, but the tag changed from `latest-staging` to `staging`.

